PR #2, opened 2018-02-06
WIP: Multi-Token Authentication & 2FA

pozylon:


pozylon:
Better late than never, this should be feature complete now, UI changes needed to support 2FA.

packages/auth/lib/challenges/SMSCodeChallenge.js
```diff
@@ -0,0 +1,89 @@
...
+    const expiresAt = new Date(new Date().getTime() + (15 * MIN_IN_MS))
+    return { payload, expiresAt }
+  },
+  isStartable: async ({ pgdb, email, user }) => {
+    return !!user.isPhoneNumberVerified
+  },
+  startChallenge: async ({ email, token, pgdb }) => {
+    const phoneNumber = await getUserPhoneNumber(pgdb, email)
+    await sendTextMessage({
+      text: `Dein Code: ${token.payload}`,
```
tpreusse: Should probably go to translations

packages/auth/graphql/schema-types.js
```diff
@@ -29,6 +30,24 @@ type User {
 
 type SignInResponse {
   phrase: String!
+  tokenTypes: [SignInTokenType]!
```
tpreusse: currently always `EMAIL_TOKEN`—rm

packages/auth/graphql/schema-types.js
```diff
@@ -10,6 +10,7 @@ type Session {
   countryFlag: String
   city: String
   isCurrent: Boolean!
+  tokenTypes: [SignInTokenType]!
```
tpreusse: rm, not really useful

packages/auth/lib/challenges/SMSCodeChallenge.js
```diff
@@ -0,0 +1,89 @@
...
+    return true
+  },
+  validateChallenge: async ({ pgdb, payload, type, user }) => {
+    console.log(`Validate SMS Code challenge for ${user.id}: ${payload} (client)`)
+    const foundToken = await pgdb.public.tokens.findOne({
+      type,
+      payload: payload.toUpperCase()
+    })
+
+    return foundToken.id
```
tpreusse: 💥— foundToken can be undefined (or null?) when the payload does not exist

packages/auth/lib/challenges/SMSCodeChallenge.js
```diff
@@ -0,0 +1,89 @@
...
+  startChallenge: async ({ email, token, pgdb }) => {
+    const phoneNumber = await getUserPhoneNumber(pgdb, email)
+    await sendTextMessage({
+      text: `Dein Code: ${token.payload}`,
+      phoneNumber
+    })
+    return true
+  },
+  validateChallenge: async ({ pgdb, payload, type, user }) => {
+    console.log(`Validate SMS Code challenge for ${user.id}: ${payload} (client)`)
```
tpreusse: rm logging of payload before going prd

packages/auth/lib/challenges/TOTPChallenge.js
```diff
@@ -0,0 +1,51 @@
...
+  },
+  startChallenge: async (options) => {
+    // no challenge transport needed, time based
+    return true
+  },
+  validateChallenge: async ({ pgdb, payload, user }) => {
+    if (!user.isTOTPChallengeSecretVerified) return false
+    const otp = OTP({ secret: user.TOTPChallengeSecret })
+    const comparablePayload = await otp.totp()
+    console.log(`Validate TOTP challenge for ${user.id}: ${comparablePayload} (server) ==? ${payload} (client)`)
```
tpreusse: rm logging of payload before going prd

packages/auth/lib/challenges/EmailTokenChallenge.js
```diff
@@ -0,0 +1,91 @@
...
+  ${geoString ? '\nLogin attempt from ' + geoString + '\n' : ''}
+  Verify that the provided security code matches *${phrase}* before proceeding.
+
+  Then please follow this link to signin.
+  ${verificationUrl}
+  `
+    })
+  },
+  validateChallenge: async ({ pgdb, payload, type, user }) => {
+    console.log(`Validate E-Mail Token challenge for ${user && user.id}: ${payload} (client)`)
```
tpreusse: rm logging of payload before going prd

packages/auth/graphql/schema.js
```diff
@@ -41,5 +52,20 @@ type mutations {
...
+  initTOTPSharedSecret: SharedSecretResponse!
+
+  # Validate the sharedSecret for the first time via token payload
+  validateTOTPSharedSecret(totp: String): Boolean!
+
+  # Activate or deactivate 2FA
+  updateTwoFactorAuthentication(enabled: Boolean!): Boolean!
+
+  # Resend Verification Code to confirm a phone number
+  resendPhoneNumberVerificationCode: Boolean!
```
tpreusse: why is this called re-send and not just send?

packages/auth/lib/Sessions.js
```diff
@@ -1,36 +1,77 @@
...
+      t."expiresAt" as "tokenExpiresAt"
+    FROM
+      sessions s
+    RIGHT OUTER JOIN
+      tokens t
+      ON t."sessionId" = s.id
+    WHERE
+      t.payload = :payload AND
+      t.type = :type AND
+      t."expiresAt" > now()
```
patte: minor: >= now()

packages/auth/lib/Sessions.js
```diff
@@ -1,36 +1,77 @@
...
 
-  return session
+const sessionByToken = async ({ pgdb, token, email: emailFromQuery, ...meta }) => {
+  const sessions = await pgdb.query(`
+    SELECT DISTINCT
+      s.*,
+      t."expiresAt" as "tokenExpiresAt"
+    FROM
+      sessions s
+    RIGHT OUTER JOIN
```
patte: omit OUTER: "The words INNER and OUTER are optional in all forms. INNER is the default; LEFT, RIGHT, and FULL imply an outer join."
https://www.postgresql.org/docs/9.2/static/queries-table-expressions.html

patte: Why do you use a RIGHT join here? Like this you get a result if the token exists, but no session for it. And you return that "half" result claiming it's a session. This should be a [INNER] JOIN.

packages/auth/graphql/resolvers/Session.js
```diff
@@ -33,5 +37,14 @@ module.exports = {
...
     return session.sid === req.sessionID
+  },
+  async tokenTypes (session, args, { pgdb }) {
+    const user = await pgdb.public.users.findOne({
+      email: session.sess.email
+    })
+    const types = await Promise.all(Object.keys(TokenTypes).filter((type) =>
+      isStartable({ type: TokenTypes[type], session, pgdb, user })
+    ))
+    return [...new Set(types)]
```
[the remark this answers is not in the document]
pozylon: I wanted to expose which 2FA are startable so that the UI can present the appropriate component on the "Authorize" page and unauthorizedSession returns a Session object. You will never get a Session object if you don't have at least one valid token challenge.

packages/auth/graphql/schema.js
```diff
@@ -15,12 +15,13 @@ type queries {
...
   # only returns users with a public profile
   user(slug: String): User
+
   # search for users
   # required role: editor
   users(search: String!, role: String!): [User]!
 
   # search for an unverified session by token
-  unauthorizedSession(email: String!, token: String!): Session
+  unauthorizedSession(email: String!, token: SignInTokenChallenge!): Session
```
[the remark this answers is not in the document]
pozylon: oh yes

packages/auth/lib/Users.js
```diff
@@ -0,0 +1,373 @@
...
+  }
+}
+
+const authorizeSession = async ({ pgdb, tokens, email: emailFromQuery, signInHooks = [] }) => {
+  // validate the challenges
+  const existingUser = await pgdb.public.users.findOne({ email: emailFromQuery })
+  const tokenTypes = []
+  let session = null
+  for (const tokenChallenge of tokens) {
+    if (tokenTypes.indexOf(tokenChallenge.type) !== -1) {
```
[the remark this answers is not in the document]
pozylon: i changed the console log to mark it a little more explicit

pozylon:
> 2fa was disabled on purpose. Maybe a return type to authorizeSession instead of Boolean that can challenge the user for the 2nd piece? This would also make startChallenge in GraphQL unnecessary—the server would just start the next challenge when the first one passes and 2fa is enabled. Unless we want to support letting the user choose which 2nd factor (phone vs. totp) but I'm a bit sceptical of the use case for that.

> initTOTPSharedSecret, validateTOTPSharedSecret, updateTwoFactorAuthentication, resendPhoneNumberVerificationCode, verifyPhoneNumber could be refactored into init2FA(type: 2FAType): SharedSecretChallange!, activate2FA(payload: String): Boolean! and disable2FA. Would be more coherent and cleanly enforce one clear 2FA method

I already did that a while ago, not having separate phone number verification and TOTP init, but it didn't feel right to me as it was very implicit and made the code harder to analyze about potential security issues. 

If you init 2FA with TOTP you will need a QR Code SVG path in the response, if you init 2FA with Twilio you don't have to present anything but server side a short message is sent to your phone. Same for the payload, one time it's a code that is generated server side and just passed to the next method and one time it's a TOTP which is generated client-side via the shared secret. It's both payloads with type string but the simplicity of the API surface would lead to more documentation needed than right now.

packages/auth/lib/Users.js
```diff
@@ -0,0 +1,372 @@
...
+  }
+}
+
+const resolveUser = async ({ slug, pgdb, fallback }) => {
+  const user = await pgdb.public.users.findOne(
+  isUUID.v4(slug)
+    ? {id: slug}
+    : {username: slug})
+  return user || fallback
+}
```
patte: remove fallback, this is not the intended behaviour. If one calls `updateEmail`,  `clearSession`, etc. and providing a  `userId` for a user which doesn't exist, the mutation must throw and not just assume me as the user. This could lead to unintended changes on me, when I would want to change another user. see [updateEmail](https://github.com/orbiting/backends/pull/2/files#diff-deae4cda0bc3ec807b15232f18604798L22)

packages/auth/lib/challenges/SMSCodeChallenge.js
```diff
@@ -0,0 +1,89 @@
...
+    throw new Error('Phone number not valid for Text Messaging')
+  }
+}
+
+module.exports = {
+  generateSharedSecret: async ({ pgdb, user, email }) => {
+    const sharedCode = generateSMSTokenCode()
+    const phoneNumber = await getUserPhoneNumber(pgdb, email)
+    await sendTextMessage({
+      text: `SMS Login Authorisierung: ${sharedCode}`,
```
patte: needs to go to translations

packages/auth/lib/challenges/SMSCodeChallenge.js
```diff
@@ -0,0 +1,89 @@
...
+  try {
+    const parsedPhoneNumber = parse(user.phoneNumber || '', 'CH') // it could be any arbitrary string
+    return format(parsedPhoneNumber.phone, parsedPhoneNumber.country, 'E.164')
+  } catch (e) {
+    throw new Error('Phone number not valid for Text Messaging')
+  }
+}
+
+module.exports = {
+  generateSharedSecret: async ({ pgdb, user, email }) => {
```
patte: where does email come from? as far as i can see [resendPhoneNumberVerificationCode.js](packages/auth/graphql/resolvers/_mutations/resendPhoneNumberVerificationCode.js) only provides user, not email.

packages/auth/lib/challenges/SMSCodeChallenge.js
```diff
@@ -0,0 +1,89 @@
...
+  return text
+}
+
+async function getUserPhoneNumber (pgdb, email) {
+  const user = await pgdb.public.users.findOne({ email })
+  try {
+    const parsedPhoneNumber = parse(user.phoneNumber || '', 'CH') // it could be any arbitrary string
+    return format(parsedPhoneNumber.phone, parsedPhoneNumber.country, 'E.164')
+  } catch (e) {
+    throw new Error('Phone number not valid for Text Messaging')
```
patte: needs to be a translated error.

packages/auth/lib/Users.js
```diff
@@ -0,0 +1,372 @@
...
+
+const updateUserTwoFactorAuthentication = async ({ pgdb, userId: id, enabled: isTwoFactorEnabled }) => {
+  const transaction = await pgdb.transactionBegin()
+  try {
+    const user = await transaction.public.users.updateAndGetOne(
```
patte: the user needs to be notified by email if two factor is disabled.

packages/auth/graphql/schema.js
```diff
@@ -41,5 +52,20 @@ type mutations {
   # if userId is null, the logged in user's sessions get cleared
   # required role to clear other's session: supporter
   clearSessions(userId: ID): Boolean!
+
+  # Generate a new sharedSecret for TOTP
+  initTOTPSharedSecret: SharedSecretResponse!
```
patte: screen shot 2018-04-04 at 13 41 29

patte:
based on our discussion:
- add user.enabledSecondFactorTokenTypes
- atomically restrict updates based on enabledSecondFactorTokenTypes
- change unauthorizedSession to unauthorizedSigninRequest(email: String!, tokenChallenge: SignInTokenChallenge!): unauthorizedSignInResponse!
  unauthorizedSignInResponse contains "availableSecondFactors" (better naming), unauthorizedSession: Session! 
- change authorizeSession(email: String!, tokenChallenge: SignInTokenChallenge!, secondFactor: SignInTokenChallenge): Boolean!
  to: authorizeSession(email: String!, tokens: [SignInToken!]!
- BONUS: expose API for admins and supporters, which token (types) have been used to authorize a session

pozylon:
```graphql
unauthorizedSession(email: String!, tokenChallenge: SignInTokenChallenge!): UnauthorizedSession

  type UnauthorizedSession {
    session: Session!
    tokenTypes: [TokenTypes!]!
  }
```

pozylon:
Can you guys please approve the schema and schema-types:

https://github.com/orbiting/backends/blob/totp/packages/auth/graphql/schema.js
https://github.com/orbiting/backends/blob/totp/packages/auth/graphql/schema-types.js

After we have locked the API surface for this PR, I can implement the rest of the features

pozylon:
@patte @tpreusse ping

pozylon:
We maybe could rename this to

"tokenTypesRequiredToAuthorize" or something

https://github.com/orbiting/backends/blob/f597f4a0695ae910e0f21df3dba55806860627cb/packages/auth/graphql/schema-types.js#L17

tpreusse:
Here my naming inputs:
```diff
- availableSecondFactorTokenTypes
+ requiredSecondFactors
- enabledSecondFactorTokenTypes
+ enabledSecondFactors
```

Did you choose not to rename `unauthorizedSession` on purpose?

Still open comment from @patte 
> re: svg(ecLevel: String = "M"): String!
> what are the possible values for ecLevel? A GraphQL enum would be nice, otherwise at least a comment is necessary.

Still open from my side:
> re: resendPhoneNumberVerificationCode
> why is this called re-send and not just send?

packages/auth/graphql/schema-types.js
```diff
@@ -31,6 +32,23 @@ type SignInResponse {
   phrase: String!
 }
 
+type SharedSecretResponse {
+  secret: String!
+  otpAuthUrl: String!
+  svg(ecLevel: String = "M"): String!
```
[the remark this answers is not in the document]
pozylon: ec_level — error correction level. One of L, M, Q, H. Default M. will do Enum

packages/auth/graphql/schema.js
pozylon: because the initial "send" is usually triggered from the "updatePhoneNumber"

pozylon:
@tpreusse answers above

> Did you choose not to rename unauthorizedSession on purpose?

Yes I found it best to keep it like that and change the result to UnauthorizedSession. Opinions?

packages/auth/lib/challenges/SMSCodeChallenge.js
pozylon: good catch

patte:
@tpreusse I'm not 100% happy with your naming input:
`requiredSecondFactors: [SignInTokenType]!`: suggests that all listed second factors have to be fulfilled for a session to be authorized. That's not the case, only one has to be completed. I would suggest: `availableSecondFactors`.
The rest is looking good to me.

packages/auth/graphql/schema.js
tpreusse: Have you considered that the phone number can also be updated with `updateMe`? I'd prefer to keep that as is, drop `updatePhoneNumber` and rename this to `sendPhoneNumberVerificationCode`. Sending on `updatePhoneNumber` feels like a unwanted side-effect. I'm fine with calling `updateMe` and `sendPhoneNumberVerificationCode` in succession in the frontend. We'll need to handle already entered phone number anyway.

pozylon:
Maybe we could use "secondFactorsSupported" on UnauthorizedSession just to bring up another naming 🗡

pozylon:
I like enabledSecondFactors, will use that. So do we have green lights for the schema now? @patte @tpreusse 

Last alternative would be returning user in UnauthorizedSession:

```graphql
type UnauthorizedSession {
  session: Session!
  User: User!
}

query unauthorizedSession(token: "asdf") {
   session: { ... }
   user: { enabledSecondFactors }
}
```

tpreusse:
Still open from my side: https://github.com/orbiting/backends/pull/2#discussion_r181987286

Ideal solution from my point of view:
```diff
-updatePhoneNumber
-resendPhoneNumberVerificationCode
+sendPhoneNumberVerificationCode
```
Drop `updatePhoneNumber` and build the phone number change logic into `updateMe`. Only send verification code with an explicit `sendPhoneNumberVerificationCode` call.

Everything else is fine from my side.

packages/auth/lib/Sessions.js
```diff
@@ -1,36 +1,77 @@
...
+  }
+  await new Promise(function (resolve, reject) {
+    req.session.save(function (error, data) {
+      if (error) {
+        return reject(new InitiateSessionError({ headers: req.headers, error }))
+      }
+      return resolve(data)
+    })
   })
+  const session = await pgdb.public.sessions.findOne({ sid: req.sessionID })
```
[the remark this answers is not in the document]
pozylon: i check for sessionID to exist now but here it is mentioned that it's set after creation of session (which happens right above)
https://www.npmjs.com/package/express-session#reqsessionid-1

packages/auth/lib/Users.js
```diff
@@ -0,0 +1,373 @@
...
+    })
+
+    // let the tokens expire
+    await transaction.public.tokens.update({
+      sessionId: session.id
+    }, {
+      updatedAt: new Date(),
+      expiresAt: new Date()
+    })
+    transaction.transactionCommit()
```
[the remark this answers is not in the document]
pozylon: good catch

packages/auth/lib/Sessions.js
pozylon: i don't know anymore but I had a good reason to use it that way, I think it was about will leave it like this until we have the tests

packages/auth/lib/Users.js
```diff
@@ -0,0 +1,372 @@
...
+      )
+    )
+  } catch (e) {
+    console.warn(`sign in hook failed in authorizeSession`, e)
+  }
+  return user
+}
+
+const upsertUserVerified = async({ pgdb, email }) => {
+  const transaction = await pgdb.transactionBegin()
```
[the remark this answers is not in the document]
pozylon: no it's not, good catch

packages/auth/lib/challenges/SMSCodeChallenge.js
pozylon: fixed

pozylon:
This just came up when solving the atomicity, if we do this:

```
-updatePhoneNumber
-resendPhoneNumberVerificationCode
+sendPhoneNumberVerificationCode
```

We lose an easy way to check if the user is allowed to update the phone number. Previously, the updatePhoneNumber call would have thrown an error in the case where 2fa is enabled with SMS. 

With moving the logic into updateMe we will introduce further semantic changes in the backends repo as right now, updateMe is defined in the republik backend but not in the other backends.

tpreusse:
@pozylon 
> We lose an easy way to check if the user is allowed to update the phone number.
How is the check any harder in republik-backend versus the auth package? `updateMe` can already manipulate phone numbers so it will need that restriction anyhow. Removing functionality from `updateMe` while bloating the API surface is not a real option for me (causes addition work in frontends).

Ping @patte

servers/republik/graphql/resolvers/_mutations/updateMe.js
```diff
@@ -199,6 +199,9 @@ module.exports = async (_, args, context) => {
         )
       }
     }
+    if (phoneNumber) {
```
tpreusse: `null` is used to unset, `undefined` to not update. Make sure to support `null`.

tpreusse: Let's open a issue for that one. Also needs string length limits. But should not be part of this PR. Otherwise it will never get merged.

patte:
@pozylon @tpreusse

updateMe / phoneNumber: I would suggest, that the auth package exports a lib function for `updateMe` to check if the change is allowed, something along the way of `isPhoneNumberEditable`. What do you think?

pozylon:
I can hopefully add the missing tests on Thursday, in the mean time you are free to adopt the UI to the new API.

patte:
@pozylon 👍looking forward to the tests!

pozylon:
Just wrote almost all integration tests but the biggest (authorizeSession) is still missing, can do it on Thursday 17th but not earlier :S @patte

PR #4, opened 2018-02-23
Fixes before stitches

tpreusse:
- activate inactive yearly if available on monthly end (on stripe webhook)
  - based on a scripted added on master: https://github.com/orbiting/backends/commit/3747a4bc4af0890d600b556ba270275b2bdd077f
- sync nl subscriptions in various places
- rm reduced membership purchase restriction
  switching from a monthly to a reduced yearly should be allowed. generally the restriction is somewhat arbitrary.

### DIFF
--- a/servers/republik/modules/crowdfundings/lib/activateYearlyMembership.js
+++ b/servers/republik/modules/crowdfundings/lib/activateYearlyMembership.js
@@ -1,0 +1,80 @@
+const moment = require('moment')
+const { ascending } = require('d3-array')
+
+module.exports = async (memberships, pgdb, dryRun) => {
+  const membershipTypeIndex = (
+    await pgdb.public.membershipTypes.findAll()
+  ).reduce(
+    (index, type) => {
+      index[type.id] = type
+      return index
+    },
+    {}
+  )
+  const packagesIndex = (
+    await pgdb.public.packages.findAll()
+  ).reduce(
+    (index, p) => {
+      index[p.id] = p
+      return index
+    },
+    {}
+  )
+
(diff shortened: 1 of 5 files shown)


servers/republik/modules/crowdfundings/lib/activateYearlyMembership.js
```diff
@@ -0,0 +1,80 @@
...
+    return packagesIndex[pledge.packageId].name !== 'ABO_GIVE' || !m.voucherCode
+  })
+  if (!inactiveMemberships.length) {
+    return
+  }
+
+  inactiveMemberships.sort((a, b) => ascending(a.sequenceNumber, b.sequenceNumber))
+
+  const electedMembership = (
+    inactiveMemberships.find(m => m.voucherCode === null) ||
```
uxengine: Is this line necessary considering `|| !m.voucherCode` in line 43?

tpreusse: yes because above it ensure that only none-give abos or redeemed give abos are considere for activation. Here it says that a reedemed abo should be used first (same logic as January activation).

uxengine: Ah, so basically a `null` voucherCode represents redeemed, while a "normal" membership would have an `undefined` voucherCode, correct? (Just trying to understand)

tpreusse: normal inactive memberships have a voucherCode string

PR #13, opened 2018-03-08
feat: audioSources

uxengine:
Kicking off backend changes for storing a reference to an audio version of an article. The current plan is to have input fields in an article's meta section for entering a URL of a previously uploaded mp3 and optionally aac and/or ogg. These may eventually turn into file upload fields.

Publikator PR: https://github.com/orbiting/publikator-frontend/pull/140
Frontend PR: https://github.com/orbiting/republik-frontend/pull/84

### DIFF
--- a/packages/documents/graphql/schema-types.js
+++ b/packages/documents/graphql/schema-types.js
@@ -16,15 +16,10 @@ type Episode {
   document: Document
 }
 
-enum AudioSourceType {
-  mp3
-  aac
-  ogg
-}
-
 type AudioSource {
-  type: AudioSourceType!
-  url: String!
+  mp3: String
+  aac: String
+  ogg: String
 }
 
 type Meta {
@@ -53,7 +48,7 @@ type Meta {
   # the id of the discussion itself
   discussionId: ID
   credits: JSON
-  audioSources: [AudioSource]
+  audioSource: AudioSource
 }
 
 # implements FileInterface
--- a/packages/documents/graphql/schema-types.js
+++ b/packages/documents/graphql/schema-types.js
@@ -16,6 +16,17 @@ type Episode {
   document: Document
 }
 
+enum AudioSourceType {
+  mp3
+  aac
+  ogg
+}
+
+type AudioSource {
+  type: AudioSourceType!
+  url: String!
+}
+
 type Meta {
   title: String
   slug: String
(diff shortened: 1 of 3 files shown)


packages/documents/lib/resolve.js
```diff
@@ -181,7 +181,8 @@ const metaFieldResolver = (meta, allDocuments = [], errors) => {
     series,
     dossier: resolver(meta.dossier),
     format: resolver(meta.format),
-    discussion: resolver(meta.discussion)
+    discussion: resolver(meta.discussion),
+    audioSources: meta.audioSources
```
patte: I think this is not needed

packages/documents/graphql/schema-types.js
```diff
@@ -42,6 +53,7 @@ type Meta {
   # the id of the discussion itself
   discussionId: ID
   credits: JSON
+  audioSources: [AudioSource]
```
patte: This is going to be interesting... if the publikator writes something to `audioSources` array that doesn't fit the `AudioSource` spec, apollo is going to explode on delivery of the document. But I think this fine.

uxengine:
@patte I changed my mind. I'm storing flat properties (audioSourceMp3, audioSourceAac, audioSourceOgg) now in the document's meta, and resolve that to an object on the query:
```
audioSource {
  mp3
  aac
  ogg
}
```

packages/documents/lib/resolve.js
```diff
@@ -177,11 +177,19 @@ const metaFieldResolver = (meta, allDocuments = [], errors) => {
...
+
   return {
     series,
     dossier: resolver(meta.dossier),
     format: resolver(meta.format),
-    discussion: resolver(meta.discussion)
+    discussion: resolver(meta.discussion),
+    audioSource
   }
 }
```
patte: because what you do here is a simple sync transform, I recomend you put it in the main Meta resolver in: servers/publikator/graphql/resolvers/Meta.js

patte: metaFieldResolver is only done on publish but audioSource should be available before. therefor and  because what you do here is a simple sync transform, I recommend you put it in the main Meta resolver in: servers/publikator/graphql/resolvers/Meta.js.

patte: Sorry the right place is `getMeta` in packages/documents/lib/meta.js
This works everywhere (publikator- and republik-api).

patte:
lgtm, merge at will

PR #14, opened 2018-03-12
admin features

patte:


servers/republik/modules/crowdfundings/graphql/resolvers/_mutations/updateUser.js
```diff
@@ -1,18 +1,39 @@
...
+  const transaction = await pgdb.transactionBegin()
+  try {
+    if (firstName || lastName || birthday || phoneNumber) {
+      await transaction.public.users.update({id: user.id}, {
+        firstName,
+        lastName,
+        birthday,
+        phoneNumber
+      }, {skipUndefined: true})
+    }
```
lukasbuenger: Question: If I passed `lastName: undefined` but `firstName: 'Luki'`, wouldn't `lastName` be set to undefined? Not really an issue atm, as the client still expects a complete particulars/address set.

servers/republik/graphql/resolvers/_mutations/updateNewsletterSubscription.js
```diff
@@ -1,11 +1,34 @@
...
+    t,
+    mail: {
+      updateNewsletterSubscription,
+      errors
+    }
+  } = context
+
+  const user = userId
+    ? await pgdb.public.users.findOne({ id: userId })
+    : me
```
lukasbuenger: So, if I enter a invalid (read non-existent) user id I just implicitly update my own prefs? Is this intentional?

PR #29, opened 2018-04-23
Assets Features for PDF

tpreusse:
New Features
- proxy pdfs (for cdn caching)
- allow to set zoom factor for url rendering
- allow to convert gifs to pngs
- supersede https://github.com/orbiting/backends/pull/3

### DIFF
--- a/packages/assets/express/frontend.js
+++ b/packages/assets/express/frontend.js
@@ -1,8 +1,9 @@
 const fetch = require('isomorphic-unfetch')
-const debug = require('debug')('assets:frontend')
 const { returnImage } = require('../lib')
 const {
-  FRONTEND_BASE_URL
+  FRONTEND_BASE_URL,
+  BASIC_AUTH_USER,
+  BASIC_AUTH_PASS
 } = process.env
 
 if (!FRONTEND_BASE_URL) {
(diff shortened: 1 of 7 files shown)


packages/assets/lib/returnImage.js
```diff
@@ -105,12 +119,6 @@ module.exports = async ({
           quality: 80
         })
       }
-
-      // update 'Content-Type'
-      res.set('Content-Type', webp
-        ? 'image/webp'
-        : mime
-      )
```
patte: did you check if Content-Type gets set automatically to mime? This is not set explicitly anymore.

tpreusse: as far as i understand it was redundant. There is already a set above if mime is detected: https://github.com/orbiting/backends/pull/29/files#diff-3dbdb44a7074aa4e3203947972d99e46R85

packages/assets/express/frontend.js
```diff
@@ -33,6 +41,7 @@ module.exports = (server) => {
       console.error('frontend fetch failed', result.url, result.status)
       return res.status(result.status).end()
     }
+    result.headers.set('Link', `<${frontendUrl}>; rel="canonical"`)
```
patte: Result is already done at this point, so this seems wired.
```
return returnImage({
      response: res,
      stream: result.body,
      headers: new Headers({
        ...headers.get().split(','),
        Link: `<${frontendUrl}>; rel="canonical"`
      }),
      options: {
        ...req.query,
        webp: !!webp
      }
    })
```
Seems to complicated as well. And I'm not sure how to get Headers in an isomorphic way. Can I suggest you just add a comment, that you add it there, so that returnImage copies is to the "real" response?

packages/assets/lib/returnImage.js
patte: 👍

PR #32, opened 2018-05-01
Docs

tpreusse:
To make it simpler to get started I've moved some env checks to later. Adapted the default DB url and updated the port in the republik readme.

See How To Run Guide: https://github.com/orbiting/docs/blob/master/guides/how-to-run.md


patte:
lovely!

PR #36, opened 2018-05-06
cluster

patte:
Enable clustering for api servers.
https://github.com/hunterloftis/throng for republik and publikator
https://github.com/orbiting/regiment for assets


servers/republik/__tests__/helpers.js
```diff
@@ -33,26 +34,28 @@ const createLocalApolloFetch = () => {
...
 }
 
 const throwOnOpenTransaction = async () => {
-  const locksEnd = await server.pgdb.query('SELECT count(*) FROM pg_stat_activity WHERE state = :state', {
+  const locksEnd = await pgDatabase().query('SELECT count(*) FROM pg_stat_activity WHERE state = :state', {
     'state': 'idle in transaction'
   })
   if (locksEnd && locksEnd[0] && locksEnd[0].count > 0) {
-    const locks = await server.pgdb.query('SELECT * FROM pg_stat_activity WHERE state = :state', {
+    const locks = await pgDatabase.query('SELECT * FROM pg_stat_activity WHERE state = :state', {
```
patrickvenetz:
```
pgDatabase.query(...)
```

should be

```
pgDatabase().query(...)
----------^
```

Assuming `pgDatabase` [is a function](https://github.com/orbiting/backends/pull/36/files#diff-cb0d8592dfc82623ed878f79c62ca6dfR47) this line will break.

packages/base/server.js
```diff
@@ -48,107 +48,103 @@ module.exports.run = (executableSchema, middlewares, t, createGraphqlContext) =>
...
+    server.enable('trust proxy')
+    // redirect to https
+    server.use((req, res, next) => {
+      if (!req.secure && (!IGNORE_SSL_HOSTNAME || req.hostname !== IGNORE_SSL_HOSTNAME)) {
+        res.redirect(`https://${req.hostname}${req.url}`)
+      }
+      return next()
+    })
+    // clear obsolete cf cookie
+    // this code was added 2018.01.19 - remove it after 2018.03.02
```
patrickvenetz: Wondering if this comment is still true and if it should be removed indeed.

Might be a dedicated MR.

patte: Out of topic, but maybe a good moment to remove this! 👍

packages/base/server.js
```diff
@@ -48,107 +48,103 @@ module.exports.run = (executableSchema, middlewares, t, createGraphqlContext) =>
...
+    }, callback)
+  } else {
+    return httpServer.listen(PORT, callback)
+  }
 }
 
-module.exports.close = () => {
+const close = () => {
   const { pubsub } = require('./lib/RedisPubSub')
   pubsub.getSubscriber().quit()
```
patrickvenetz: Found `quit()` to return a Promise. Would "await" it.

```
await pubsub.getSubscriber().quit()
```

packages/base/server.js
```diff
@@ -48,107 +48,103 @@ module.exports.run = (executableSchema, middlewares, t, createGraphqlContext) =>
...
   pubsub.getPublisher().quit()
```
patrickvenetz: Would await this call aswell as it returns a Promise.

packages/base/server.js
```diff
@@ -48,107 +48,103 @@ module.exports.run = (executableSchema, middlewares, t, createGraphqlContext) =>
...
-  if (subscriptionServer) {
-    subscriptionServer.close()
-  }
-  httpServer.close()
-  pgdb.close()
+  subscriptionServer && subscriptionServer.close()
+  httpServer && httpServer.close()
+  pgdb && pgdb.close()
```
patrickvenetz: `pgdb.close()`  returns a promise. Would wait for that one to.

servers/assets/index.js
```diff
@@ -1,3 +1,25 @@
...
+if (CLUSTER) {
+  server.addMiddleware(
+    (server) =>
+      server.use(regiment.middleware.MemoryFootprint(WEB_MEMORY))
+  )
+  regiment(
+    server.start,
+    {
+      numWorkers: WEB_CONCURRENCY,
+      deadline: 30000
```
patrickvenetz: Annotate units here.

I find this pattern pretty neat, assuming `deadline` is in miliseconds:

```
 deadline: 30 * 1000 // 30 seconds
```

Or if it were 5 minutes:

```
 deadline: 5 * 60 * 1000
```

Or 9 hours...

```
 deadline: 9 * 60 * 60 * 1000
```

servers/assets/index.js
```diff
@@ -1,3 +1,25 @@
 require('@orbiting/backend-modules-env').config()
 const server = require('./server')
-server.run()
+const regiment = require('regiment')
+
+const {
+  CLUSTER,
```
patrickvenetz: Maybe default `CLUSTER` to `false`

servers/assets/server.js
```diff
@@ -33,16 +40,32 @@ module.exports.run = () => {
...
 
-module.exports.close = () => {
-  server.close()
+module.exports = {
+  start,
+  close,
+  addMiddleware
 }
+
+process.on('SIGTERM', () => {
```
patrickvenetz: Maybe log that process got `SIGTERM` signal. Could be useful for post mortem analysis.

servers/publikator/index.js
```diff
@@ -1,3 +1,19 @@
...
+const {
+  CLUSTER,
+  WEB_CONCURRENCY = 1
+} = process.env
+
+if (CLUSTER) {
+  throng({
+    workers: WEB_CONCURRENCY,
+    grace: 30000,
+    start: server.run,
```
patrickvenetz: `start`, `master` receive a worker ID as an argument. Might be useful for debuging purposes. If possible, even have listeners (on `SIGTERM`) able to log worker ID.

[A Complex example](https://github.com/hunterloftis/throng#a-complex-example) looks pretty much like that.

packages/base/server.js
patte: I'm against awaiting in close, because it would leave a false impression. The close call comes from a os interrupt, which is asynchronous in it's nature and doesn't care what we do on our end. In the end the process gets killed no matter if we finished closing or not.

PR #37, opened 2018-05-07
fix(auth): Encode/decode email in verification URL

patrickvenetz:
Ensures `email` query parameter in verification URL is encoded as base64 in an url-safe manner and ensures `emailFromQuery` in `Session::sessionByToken()` is a plain string.

URLs containin `email=<email>` or `email=<urlencoded email>` instead of `email=<base64-encoded email>` let to double-encoding conflicts in some email clients, and resulted in URLs being invalid:

`heidi@wiese.tld` is encoded to `heidi%40wiese.tld` (`@` => `%40`). Some email clients would re-encode and generate a link with `heidi%2540wiese.tld` (`%` => `%25`).

⚠️ This is a breaking change and requires https://github.com/orbiting/republik-frontend/pull/101 ~to be merged an deployed first.~ Deployed.

Introduces [urlsafe-base64](https://www.npmjs.com/package/urlsafe-base64) as a dependency.

### DIFF
--- a/packages/auth/lib/Sessions.js
+++ b/packages/auth/lib/Sessions.js
@@ -1,5 +1,5 @@
 const { NoSessionError, QueryEmailMismatchError, DestroySessionError } = require('./errors')
-const urlsafeBase64 = require('urlsafe-base64')
+const { decode, match } = require('@orbiting/backend-modules-base64u')
 
 const destroySession = async (req) => {
   return new Promise((resolve, reject) => {
@@ -18,10 +18,8 @@ const sessionByToken = async ({ pgdb, token, email: emailFromQuery, ...meta }) =
     'sess @>': { token }
   })
 
-  if (urlsafeBase64.validate(emailFromQuery)) {
-    emailFromQuery =
-      urlsafeBase64.decode(Buffer.from(emailFromQuery)).toString()
-  }
+  // Recognize and decode urlsafe base64 encoded email address.
+  if (match(emailFromQuery)) emailFromQuery = decode(emailFromQuery)
 
   if (!session) {
     throw new NoSessionError({ token, emailFromQuery, ...meta })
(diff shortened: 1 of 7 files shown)


patte:
this looks beautiful: https://github.com/orbiting/republik-frontend/pull/101/files
Not just because this is auth related "lava-ground" I think both the FE and BE implementation should be the same. I see two options, copying lib/utils/base64u.js and care about idempotent orbiting npm packages later, or invest in them now. What do you think?

patrickvenetz:
@patte Thought about [it](https://github.com/orbiting/republik-frontend/pull/101/files) a bit more and find `base64u` not to be in a good, healthy, stable state yet to be published.

I do not fully comprehent what [these implemented work-around methods](https://developer.mozilla.org/en-US/docs/Web/API/WindowBase64/Base64_encoding_and_decoding#The_Unicode_Problem) do.

But fair enough: It is not likely to change much overtime as it has a single purpose, does that one for us pretty neatly.

>  I see two options, copying lib/utils/base64u.js and care about idempotent orbiting npm packages later, or invest in them now.

Hence going for first option.

Bäm.

patrickvenetz:
@patte Please revisit.

patte:
@patrickvenetz please check the failing travis tests:
`Error: Cannot find module '/home/travis/build/orbiting/backends/servers/publikator/?(__tests__|lib|graphql)/**/*.test.js'`

patrickvenetz:
Did revert installation of `faucet` as a dependency in base64u package. Failing TravisCI checks did recover.

faucet comes with [ancient a tape dependency](https://github.com/substack/faucet/blob/master/package.json).

A combination of yarn's resolution, linking of packages and `servers/republik` script `test:prepare` is removing and symlinking a node_module folder](https://github.com/orbiting/backends/blob/encode-emails-urlsafe/servers/republik/package.json#L56) leads to have `.bin/tape` to be not a recent version but the faucet's ancient one.

That ancient one does not support globbing (e.g. `**/*.js`) which let to failing TravisCI tests.

💥 

Good news: This can be straightened, with love and dedication (declaring packages, removing outside-the-box-attempts), but would like to not do so in this Merge Request.

PR #41, opened 2018-05-09
fix(auth): Remove email base64 decoding

patrickvenetz:
This Merge Requets reverts "fail safe" about having a decoded `emailFromQuery` as email value passed is to be expected an email address, not a potentially base64, URL-safe encoded email address.


patrickvenetz:
Ship it

PR #44, opened 2018-05-18
feat(github): refactoring commits endpoint

lukasbuenger:
###### Breaking API Changes

* The `Repo.commits` resolver acts as an actual paginated commit history
   * Paginated in the spirit of our usual pagination logic; count, cursors and stuff, no edges though.
   * Cursors don't point at actual commits, but at timestamps until (or since) which you want to retrieve the first n commits. This model works fine for now, but **it is impossible to determinately reconstruct a query based on timestamped cursors**.  

###### New

* Subquery `Document.children` that enables retrieval of paginated root MDAST nodes. Used by orbiting/republik-frontend#126. Currently only works for documents that have only teasers as top-level children. 

###### Under the hood

* The `Repo.commits` resolver aggregates results of [history queries](https://developer.github.com/v4/object/commit/#history) on each head of [repo/refs](https://developer.github.com/v4/object/repository#refs) in one query, skipping the Github REST API calls entirely. 
* The `Repo.commit(id)` resolver acts directly on [Repo.object](https://developer.github.com/v4/object/repository/#object), skipping Github REST calls entirely. 
* The `Repo.latestCommit` resolver is an alias of `Repo.commits(first: 1)[0]` and again, skips Github REST calls entirely. 
* The only endpoint checking for Redis versions is `Repo.commit(id)`. Not even sure if that is necessary.
* Refactored and renamed normalizer functions for commits.
   * `commitFromRest` for Github REST responses
   * `commitFromGQL` for Github GraphQL API responses
* All queries should operate on the Github GraphQL API only.


servers/publikator/lib/github/index.js
```diff
@@ -132,22 +145,156 @@ module.exports = {
...
+      .map(({ target }) =>
+        target.history.nodes
+          .map(
+            commit => commitFromGQL(repo, commit)
+          )
+      )
+      .reduce(
+        (acc, v) => acc.concat(v), []
+      )
+      .filter((v, i, arr) => arr.map(mapObj => mapObj.id).indexOf(v.id) === i)
```
tpreusse: use `arr.findIndex(node => node.id === v.id)`

servers/publikator/lib/github/index.js
```diff
@@ -132,22 +145,156 @@ module.exports = {
...
+      .sort((a, b) => descending(a.date, b.date))
+      .slice(0, first)
```
tpreusse: If this slices here it potentially needs to change `hasNextPage` to true and set a proper `endCursor` (e.g. if first is 100 and no single branch had more than 100 commits, but 3 had 50)

servers/publikator/lib/github/index.js
```diff
@@ -132,22 +145,156 @@ module.exports = {
...
+                }
+              }
+            }
+          }
+        }
+      `,
+      variables: {
+        login,
+        repoName,
+        maxRefs: 100,
```
tpreusse: what will happen once magazine front hit 100 branches? (currently at 95)

servers/publikator/lib/github/index.js
```diff
@@ -23,7 +24,7 @@ const tagNormalizer = (tag, repoId, refName) => ({
   }
 })
 
-const commitNormalizer = ({
+const commitFromRest = ({
```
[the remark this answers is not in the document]
tpreusse: or get ride of `commitFromRest` if no longer used, and name the new one `commitNormalizer`.

servers/publikator/lib/github/index.js
```diff
@@ -132,22 +145,156 @@ module.exports = {
...
+        (acc, v) => acc.concat(v), []
+      )
+      .filter((v, i, arr) => arr.map(mapObj => mapObj.id).indexOf(v.id) === i)
+      .sort((a, b) => descending(a.date, b.date))
+      .slice(0, first)
+
+    return {
+      pageInfo: {
+        endCursor: (commits.length && commits.slice(-1)[0].date) || null,
+        startCursor: (commits.length && commits[0].date) || null,
```
tpreusse: when querying with those cursor, will github include the first/last item again or will it be excluded? Does that mean the fundamental limitations is that when two commits share the same timestamp and occur around the pagination edge one could be missed?

servers/publikator/graphql/resolvers/Repo.js
```diff
@@ -16,47 +13,10 @@ const debug = require('debug')('publikator:repo')
...
-      .then(refs => refs
-        .map(ref => ref.target)
-        .sort((a, b) => descending(a.author.date, b.author.date))
-        .shift()
-      )
-      .then(({ oid: sha }) =>
-        getCommit(repo, { id: sha }, context)
-      )
+    return getCommits(repo, { first: 1 }, context)
+      .then(commitConnection => commitConnection.nodes.shift())
```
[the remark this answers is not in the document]
lukasbuenger: True, on the other hand you miss out on a null check. `slice(0, 1)`?

servers/publikator/lib/github/index.js
lukasbuenger: This is actually just ported from the previous implementation. Can we support more than 100 branches? Can we do it in the UI?

servers/publikator/lib/github/index.js
lukasbuenger: True

servers/publikator/lib/github/index.js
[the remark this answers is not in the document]
lukasbuenger: We cannot get rid of the rest normalization method as mutations still use the REST interface. 

As for the naming, can do.

servers/publikator/lib/github/index.js
lukasbuenger: Both `since` and `until` are piped through to [git-log](https://git-scm.com/docs/git-log) somehow and both work by excluding the items with the exact equality on the timestamp, therefore it can even get used as cursor.  
As for the pagination corner case: It is not possible to have two commits added to the same repo at the same time in git. The same has got to be true for Github. The only scenario, where it has to handle possible timestamp equality is when syncing local and remote heads, which we never do. 
At least, that was my reasoning behind it.

lukasbuenger:
Hey, many thanks @tpreusse! See comments above.

lukasbuenger:
The tests are not green because they correctly aren't passing.

servers/publikator/graphql/resolvers/Repo.js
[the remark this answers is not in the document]
tpreusse: Which null check?
```
[].shift()
undefined
[][0]
undefined

null.shift()
Uncaught TypeError: Cannot read property 'shift' of null
null[0]
Uncaught TypeError: Cannot read property '0' of null
```

patte:
hey @lukasbuenger why are the tests not passing, I think they should.

PR #47, opened 2018-05-22
maintenance: update dependencies

patte:
TODOs
- [x] check all major upgrades and apply if reasonable


tpreusse:
What's the reason to already go to node 10? It will only become `LTS` in October. I'd wait a bit more for it to prove stable and adapt it once it ships with npm 6 and get ride of yarn :P

PR #48, opened 2018-05-23
fix(Statements): obey correct order of users

tpreusse:
before the `focus` (internally called `firstId`) came in a random spot
screen shot 2018-05-23 at 11 51 22

afterwards it is guaranteed that it comes first
screen shot 2018-05-23 at 11 51 35


tpreusse:
This fixed the statement direct links: https://www.republik.ch/community?id=462d5d87-ed71-4868-beb7-77203cd16012 (currently a random first result is highlighted)

PR #49, opened 2018-05-24
fix(auth): Harden querying, resolving user with slug, userId

patrickvenetz:


packages/auth/lib/Users.js
```diff
@@ -293,12 +293,19 @@ const upsertUserAndConsents = async({ pgdb, email, consents, req }) => {
 }
 
 const resolveUser = async ({ slug, pgdb, userId }) => {
-  const idOrSlug = slug || userId
-  const user = await pgdb.public.users.findOne(
-  isUUID.v4(idOrSlug)
-    ? {id: idOrSlug}
-    : {username: idOrSlug})
-  return user || pgdb.public.users.findOne({ id: userId })
```
tpreusse: Since `slug: false` is not a real case I'd prefer to revert to the old one but also remove the second query (`|| pgdb.public.users.findOne({ id: userId })`)

patrickvenetz: Done so.

PR #50, opened 2018-05-24
fix updateMe

patte:


packages/auth/lib/Users.js
```diff
@@ -390,25 +391,26 @@ const updateUserEmail = async ({ pgdb, user, email }) => {
 }
 
 const updateUserPhoneNumber = async ({ pgdb, userId, phoneNumber }) => {
-  const user = pgdb.public.users.findOne({ id: userId })
+  const user = await pgdb.public.users.findOne({ id: userId })
```
patrickvenetz: 😱

packages/auth/lib/Users.js
```diff
@@ -390,25 +391,26 @@ const updateUserEmail = async ({ pgdb, user, email }) => {
...
+    throw new TranslatedError(t('api/auth/sms/phone-number-not-valid'))
   }
 
   try {
     return pgdb.public.users.updateAndGetOne(
       {
         id: userId
       }, {
-        phoneNumber,
+        phoneNumber: formattedPhoneNumber,
```
patrickvenetz: `formattedPhoneNumber` may be falsy on purpose?

PR #52, opened 2018-05-25
split basic auth config in assets package

patte:
set BASIC_AUTH_USER, BASIC_AUTH_PASS to secure the api itself.
set FRONTEND_BASIC_AUTH_USER, FRONTEND_BASIC_AUTH_PASS to authenticate to the frontend (/render, /frontend)

### DIFF
--- a/packages/assets/README.md
+++ b/packages/assets/README.md
@@ -24,7 +24,7 @@ Check [assets-backend](https://github.com/orbiting/assets-backend) for a deploya
 
   renders :url with a viewport of :width x :hight to a png. webp conversion not supported.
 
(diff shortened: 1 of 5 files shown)


patrickvenetz:
Why is this?

patte:
Before this split. If you set BASIC_AUTH_USER and PASS, two things happened:
- the api itself (all routes) were protected by basic auth
- requests to the FRONTEND_URL (for /frontend and /render) were authenticated with the same basic auth credentials
Now, these two things are separate, which is more practical for staging. We want the api not to be protected (otherwise visitors must enter the basic auth credentials twice, first for the frontend and second for the api server), but calls to the protected frontend must be authenticated to work. Yes this adds a backdoor to the protected frontend via the unprotected backend, but we don't do secret things in staging anyway.

patrickvenetz:
Got it. 👍

PR #53, opened 2018-06-05
App

tpreusse:
Mostly doing this to create a review app ;)

edit by @patte
TODOs:
- [x] send notifications for discussion events

not in this PR:
- [ ] tests for app notifications (upsertDevice, rollDeviceToken, removeDevice)
- [ ] move republik discussion into it's own package
- [ ] move `servers/republik/lib/Notifications` into notifications package (depends on the task above)


patrickvenetz:
Over-all wondering if we should keep a record of messages sent and if a message was dispatched successfully delivered if Firebase and Apple Push Notification Service do report that.

Apart from debugging I believe that would come in handy for Careteam purposes as well as some stats.

(If data can be retrieved on either providers end, good too.)

packages/notifications/lib/firebase.js
```diff
@@ -0,0 +1,71 @@
+// https://firebase.google.com/docs/cloud-messaging/admin/legacy-fcm?authuser=1
+// https://firebase.google.com/docs/cloud-messaging/admin/send-messages?authuser=1#defining_the_message
+// https://developer.apple.com/library/archive/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/CommunicatingwithAPNs.html
```
patrickvenetz: Move this comment to `lib/apn.js`

packages/notifications/graphql/resolvers/Device.js
```diff
@@ -0,0 +1,17 @@
...
+module.exports = {
+  async user (device, args, { pgdb, user: me }) {
+    const user = await pgdb.public.users.findOne({
+      id: device.userId
+    })
+      .then(user => transformUser(user))
+    Roles.ensureUserIsMeOrInRoles(user, me, userAccessRoles)
+    return user
+  },
+  lastSeen (device) {
```
patrickvenetz: Naming: Not sure if `updatedAt` really represents a last seen timestamp.

I see `updatedAt` updated when device is added or a (device) token is rolled.

I'd feel fine with `lastSeen` if timestamp is updated everything a request from a device comes in (it does not update, does it?).

packages/notifications/lib/app.js
```diff
@@ -0,0 +1,33 @@
+const { publish: publishFirebase } = require('./firebase')
+const { publish: publishiOS } = require('./apn')
+const debug = require('debug')('notifications:publish')
+
+const defaultTTL = 4 * 7 * 24 * 60 * 60 * 1000 // 4 weeks in ms, firebase default
```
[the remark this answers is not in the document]
patte: I always do the opposit, as I think it's more intuitive to read: 4 weeks * 7 days * 24 hours, gives you hours.

packages/notifications/migrations/sqls/20180628154012-devices-up.sql
```diff
@@ -0,0 +1,13 @@
+CREATE TABLE "devices" (
+  "id"            uuid primary key not null default uuid_generate_v4(),
+  "userId"        uuid not null references users(id),
+  "sessionId"     uuid not null unique references sessions(id) ON UPDATE CASCADE ON DELETE CASCADE,
+  "token"         text unique,
+  "information"   jsonb,
```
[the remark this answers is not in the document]
patte: I would have liked `info` the most, as it's device information. The only one here really. But because the App would need to adapt, I think we should let it as it is.

packages/notifications/graphql/resolvers/Device.js
patte: > (it does not update, does it?)

no it doesn't on every request. But the app is supposed to call `upsertDevice` everytime it comes back from the background. So it's still fairly up to date. Do you think that's ok?

patrickvenetz: I think that is very much so ok.

packages/auth/lib/Users.js
```diff
@@ -40,9 +41,91 @@ const {
...
+      }
+      // undefined order for others
+      return 0
+    })
+}
+
+const setPreferredFirstFactor = async (user, tokenType = null, pgdb) => {
+  const enabledTokenTypes = await enabledFirstFactors(user._raw.email, pgdb)
+  if (tokenType && enabledTokenTypes.indexOf(tokenType) === -1) {
+    throw new TokenTypeNotEnabledError({ tokenType })
```
tpreusse: Always allow to set preference to known token types ('APP', 'EMAIL'). FE will relax UI restrictions, letting users set the app preference independently of whether a device is available at the exact moment.

PR #54, opened 2018-06-07
fix(mail): Query with lower-cased, hashed email address

patrickvenetz:
MailChimp stores, however added, email addresses all lower-cased. This Pull Request makes sure `MailchimpInterface` is querying MailChimp API with a MD5-hashed, _lower-cased_ email address.

It fixes a bug in which we did query API with a case-sensitive email address and did not receive any results.

### DIFF
--- a/packages/mail/MailchimpInterface.js
+++ b/packages/mail/MailchimpInterface.js
@@ -26,7 +26,6 @@ const MailchimpInterface = ({ logger }) => {
         .createHash('md5')
         .update(email.toLowerCase())
         .digest('hex')
-        .toLowerCase()
 
       return this.buildApiUrl(`/members/${hash}`)
     },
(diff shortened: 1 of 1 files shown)


tpreusse:
Check for other occurances. Had to fix the same issue in a different place 19 days ago: https://github.com/orbiting/backends/commit/ad53c2c13f146e537d0992669e45f137caa151df

packages/mail/MailchimpInterface.js
```diff
@@ -24,7 +24,7 @@ const MailchimpInterface = ({ logger }) => {
     buildMembersApiUrl (email) {
       const hash = crypto
         .createHash('md5')
-        .update(email)
+        .update(email.toLowerCase())
         .digest('hex')
         .toLowerCase()
```
tpreusse: afaik this is wrong, no need to lower case the md5

patrickvenetz: Removed.

patrickvenetz:
I believe all requests handeled in `mail` in package are piped through that interface. Email addresses in API request bodies remain untouched.

PR #59, opened 2018-06-22
Fix: Sending preview email did not work

patrickvenetz:
A preview newsletter is a `Document` in ElasticSearch, having `meta.template:editorialNewsletter` set but not in feed (`meta.feed:false`). Default search query does not return newsletters.

`getDocument` can't, and should neither be allowed, to retrieve preview email via `repoId`.

This Pull Request changes [Document filter query to allow returning preview newsletter](dd56238), set in environment variable `PREVIEW_MAIL_REPO_ID`.

It does [functionalize `filters`](51fec56) (like `rolebasedFilter` is already)

### DIFF
--- a/packages/search/lib/indices/comments.js
+++ b/packages/search/lib/indices/comments.js
@@ -19,18 +19,20 @@ module.exports = {
       'resolved.user.facebookId': {}
     },
     filter: {
-      bool: {
-        must: [
-          { term: { __type: type } },
-          { term: { published: true } },
-          { term: { adminUnpublished: false } }
-        ],
-        must_not: [
-          { terms: { discussionId: [
-            '3c625fe4-788f-44d5-ad5e-ac93bd9a6292' // Crowdfunding discussions
-          ] } }
-        ]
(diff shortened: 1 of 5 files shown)


packages/search/lib/indices/comments.js
```diff
@@ -19,18 +19,20 @@ module.exports = {
...
+      default: () => ({
+        bool: {
+          must: [
+            { term: { __type: type } },
+            { term: { published: true } },
+            { term: { adminUnpublished: false } }
+          ],
+          must_not: [
+            { terms: { discussionId: [
+              '3c625fe4-788f-44d5-ad5e-ac93bd9a6292' // Crowdfunding discussions
```
patte: off topic, but only realised now. This static ID is not so pretty to have in the code. How about `SEARCH_EXCLUDE_DISCUSSION_IDS` to configure dynamically?

patrickvenetz: It is a good idea. Will follow.

PR #67, opened 2018-07-04
Value-Added-Tax for package options

patrickvenetz:
This Pull Request adds a `vat` attribute to `packageOptions` table. It's value is an integer. 100 would represent 1%. Default value is 0.

When creating a pledge and it's options, VAT value of package option is copied into `vat` attribute in `pledgeOptions` table.

It further exposes VAT value in GraphQL `PackageOption` type.

This enables to create invoices (upcoming changes), taking VAT into account.

A migration for existing pledge options is run manually.

### DIFF
--- a/servers/republik/migrations/20180703153921-package-options-vat.js
+++ b/servers/republik/migrations/20180703153921-package-options-vat.js
@@ -1,0 +1,47 @@
+'use strict'
+
+var fs = require('fs')
+var path = require('path')
+var Promise
+
+/**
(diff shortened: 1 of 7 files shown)


patte:
So this is just for representation purposes. No calculations are done with it, because our prices always include VAT already, right?
Could you post the migration for the existing pledgeOptions here?

patrickvenetz:
> So this is just for representation purposes. No calculations are done with it, because our prices always include VAT already, right?

Yes.

> Could you post the migration for the existing pledgeOptions here?

Yes.

PR #71, opened 2018-07-09
fix(search): Invalidate cache when scheduling a document to publish

patrickvenetz:
Cache should also be invalidated when a document is scheduled to publish. So far cache got only invalidated when a document's status was supposed to change it's `__state` to published, prepublished.

### DIFF
--- a/packages/search/lib/Documents.js
+++ b/packages/search/lib/Documents.js
@@ -514,6 +514,8 @@ const publishScheduled = (elastic, redis, elasticDoc) => ({
       elasticDoc.meta.repoId,
       elasticDoc.id
     )
+
+    await createCache(redis).invalidate()
   },
   afterScheduled: async () => {
     await elastic.update({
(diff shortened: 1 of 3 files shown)


packages/search/lib/Documents.js
```diff
@@ -514,6 +514,8 @@ const publishScheduled = (elastic, redis, elasticDoc) => ({
       elasticDoc.meta.repoId,
       elasticDoc.id
     )
+
+    await createCache(redis).invalidate()
```
patte: so, this is not needed anymore, isn't it?

packages/search/lib/Documents.js
```diff
@@ -567,6 +569,8 @@ const prepublishScheduled = (elastic, redis, elasticDoc) => ({
...
```
patte: so, this is not needed anymore, isn't it?

patrickvenetz: Yes.

PR #72, opened 2018-07-09
Automatically cancel monthly memberships if upgraded to yearly 

lukasbuenger:
This was successfully tested locally and with manually mocked Stripe responses. Will not merge, unless I can test this against the actual workflow, including Stripe webhook events.

### DIFF
--- a/servers/republik/modules/crowdfundings/lib/generateMemberships.js
+++ b/servers/republik/modules/crowdfundings/lib/generateMemberships.js
@@ -1,5 +1,6 @@
 const moment = require('moment')
 const { getPledgeOptionsTree } = require('./Pledge')
+const cancelMembership = require('../graphql/resolvers/_mutations/cancelMembership')
 const debug = require('debug')('crowdfundings:memberships')
 const { enforceSubscriptions } = require('./Mail')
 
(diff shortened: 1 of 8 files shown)


servers/republik/modules/crowdfundings/lib/generateMemberships.js
```diff
@@ -68,6 +76,12 @@ module.exports = async (pledgeId, pgdb, t, logger = console) => {
             membership
           }
         } else {
+          if (
+            ['ABO_BENEFACTOR', 'ABO', 'ABO_REDUCED'].includes(membershipType.name) &&
```
patrickvenetz: `ABO_BENEFACTOR`  and `ABO_REDUCED` are non-existant "membership types". Terms exists as made-up terms for CSV export to distinguish data. There is probably just a `BENEFACTOR_ABO`, but no reduced version of it.

Checking up on that.

patrickvenetz:
@lukasbuenger Pushed a slightly revised version. Would appreciate if you'll have a more fond peak. This Pull Request is scoped to cancel any active subscription (`MONTHLY_ABO` for now) if there an annual membership is bought.

patrickvenetz:
I believe this can be deployed without https://github.com/orbiting/republik-frontend/pull/138 and I'll have a go at that one later this week.

patrickvenetz:
> Does buying an ABO on top of a MONTHLY_ABO not trigger the new behaviour?

Yes, buying an ABO, or BENEFACTOR_ABO, on top of a MONTHLY_ABO will cancel MONTHLY_ABO.

Did post a notice to Careteam already.

PR #74, opened 2018-07-10
fix(republik): Bubble focused comment with parents to very top of tree

patrickvenetz:
This Pull Request ensures that a focused comment are put on top of a comment tree.

Before this Pull Request, top-level-comments with children did not bubble to the top but got focus attributes (e.g. background color) and were scrolled to. Yet, if hidden behind pagination horizon, got not shown at all.


patrickvenetz:
(Broken tests unrelated to this Pull Request. Pick-up pieces hereafter.)

PR #76, opened 2018-07-17
feat(auth): TTL of default session a year, shorter for special roles

patrickvenetz:


packages/auth/express/auth.js
```diff
@@ -67,13 +72,26 @@ exports.configure = ({
...
   // Initialise Passport
   server.use(passport.initialize())
   server.use(passport.session())
+
+  server.use(function (req, res, next) {
+    // Check if a user has more than one and let session expire after a shorter period of time
+    if (req.user && userIsInRoles(req.user, specialRoles)) {
+      req.session.cookie.expires =
+        new Date(Date.now() + SHORT_MAX_AGE_IN_MS)
+      req.session.cookie.maxAge = SHORT_MAX_AGE_IN_MS
```
patte: 1. ~You have to save the session, to be sure this is applied. Maybe you could extract [this](https://github.com/orbiting/backends/blob/app-authenticator/packages/auth/lib/Sessions.js#L35) into a lib and use it here.~
2. ~You basically added another rolling mechanism here (resulting in +1 database queries). You should check if maxAge is not correct and only modify it when needed. I tested this and it doesn't work, because as soon as this middleware doesn't set the maxAge/expire, the rolling mechanism does, and it reset's the expire based on the configured maxAge. I'm not sure what's the best approach. Maybe keep this middleware but disable rolling.~
Never mind, setting it here, without saving, makes the rolling machanism of express-session use your `maxAge`. So everything should be fine.
https://github.com/expressjs/session/blob/master/session/cookie.js#L39

packages/auth/express/auth.js
```diff
@@ -1,8 +1,13 @@
...
-const PgSession = require('connect-pg-simple')(session)
 const passport = require('passport')
+const PgSession = require('connect-pg-simple')(session)
+
 const transformUser = require('../lib/transformUser')
 const basicAuthMiddleware = require('./basicAuth')
+const { specialRoles, userIsInRoles } = require('../lib/Roles')
+
+const DEFAULT_MAX_AGE_IN_MS = 60000 * 60 * 24 * 365 // 1 year
+const SHORT_MAX_AGE_IN_MS = 60000 * 60 * 24 * 1 // 1 week
```
patte: paranoid me would do the opposite:
`LONG_MAX_AGE_IN_MS= 1 year`
`DEFAULT_MAX_AGE_IN_MS = 1 week`
but your solution is ok aswell.

packages/auth/express/auth.js
```diff
@@ -67,13 +72,26 @@ exports.configure = ({
...
   })
 
   // Initialise Passport
   server.use(passport.initialize())
   server.use(passport.session())
+
+  server.use(function (req, res, next) {
+    // Check if a user has more than one and let session expire after a shorter period of time
+    if (req.user && userIsInRoles(req.user, specialRoles)) {
+      req.session.cookie.expires =
```
patte: I think setting maxAge is enough:
- https://github.com/expressjs/session#rolling
- https://github.com/voxpelli/node-connect-pg-simple/blob/master/index.js#L207

P.S. stupid prettier making two lines out of this :1st_place_medal:

patrickvenetz: That worked. To set `maxAge` is enough.

packages/auth/express/auth.js
patrickvenetz: Not giving in to paranoia here.

patrickvenetz: (To be fair: I did have the very same thought. But overcame my bias.)

packages/auth/express/auth.js
```diff
@@ -67,13 +72,26 @@ exports.configure = ({
...
+
     next(null, user)
   })
 
   // Initialise Passport
   server.use(passport.initialize())
   server.use(passport.session())
+
+  server.use(function (req, res, next) {
+    // Check if a user has more than one and let session expire after a shorter period of time
```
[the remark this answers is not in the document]
patrickvenetz: Fixed.

PR #77, opened 2018-07-17
feat(republik): Add invoice related fields to companies

patrickvenetz:
This Pull Request is a mini-migration. It adds free-text fields required to generate invoices:
- An (invoice sender) address
- A VAT Identification Number
- Bank details (e.g. IBAN)
- An invoice numbering prefix (e.g. `FOO-123123`)

When generating invoices, data if these files is copied to in to (immutable) invoice tables. While this company invoice information may be stored in translation files, putting attaching them to `companies` in database seemed more strict.

### DIFF
--- a/servers/republik/migrations/20180717161308-company-details.js
+++ b/servers/republik/migrations/20180717161308-company-details.js
@@ -1,0 +1,47 @@
+'use strict'
+
+var fs = require('fs')
+var path = require('path')
+var Promise
+
+/**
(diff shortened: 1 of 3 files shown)


patte:
I totally approve the code changes in this PR, but I would prefer if you would bundle features in PRs. Why not add all the rest of the changes needed for invoices into this PR? Then they can be reviewed as a hole and the danger of dead code of unfinished features is mitigated.

patrickvenetz:
True, that was a little too atomic.

PR #78, opened 2018-07-31
feat(search): Create search tracking index per day

patrickvenetz:
Creates search tracking indices with this pattern: `republik-searches-<YYYY-mm-dd>`. Data can be aggregated with wildcarded index queries like `republik-searches-*` (or via Kibana index patterns).

This prevents indices to grow too big to handle and allows to remove data over time more easily by removing a date range.

It also adds aggregation results, and user roles. Latter should allow to separate internal traffic from external.

### DIFF
--- a/packages/search/lib/utils.js
+++ b/packages/search/lib/utils.js
@@ -6,8 +6,15 @@ const ES_INDEX_PREFIX = 'republik' // TODO: Put into config
  * @example "2018-05-24-17-05-23"
  * @return {String} [description]
  */
-const sanitizedDate =
+const getDateTime =
   () => new Date().toISOString().slice(0, -5).replace(/[^\d]/g, '-')
+
+/**
+ * @example "2018-05-24"
+ * @return {String} [description]
+ */
+const getDate =
+  () => new Date().toISOString().slice(0, -14).replace(/[^\d]/g, '-')
 
 /**
  * @param  {String} name An index name without prefix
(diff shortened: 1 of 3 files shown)


packages/search/lib/utils.js
```diff
@@ -6,9 +6,16 @@ const ES_INDEX_PREFIX = 'republik' // TODO: Put into config
...
-const sanitizedDate =
+const getDateTime =
   () => new Date().toISOString().slice(0, -5).replace(/[^\d]/g, '-')
 
+/**
+ * @example "2018-05-24"
+ * @return {String} [description]
+ */
+const getDate =
+  () => new Date().toISOString().slice(0, -14).replace(/[^\d]/g, '-')
```
patte: you know we use [moment](https://github.com/orbiting/backends/blob/b2818c039254f5aac9dd6c166524fd018317b9c0/servers/republik/package.json#L29) in other places already don't you? Anyway I'm fine with this as well.

packages/search/lib/utils.js
```diff
@@ -21,8 +28,15 @@ const getIndexAlias =
  * @param  {String} name An index name without prefix
  * @return {String}      An index name with a date timestamp
  */
-const getIndexDated =
-  (name) => [ES_INDEX_PREFIX, name, sanitizedDate()].filter(Boolean).join('-')
+const getDatetimeIndex =
```
patte: I would like `getDateTimeIndex` better, as it's camelCase everywhere else.

packages/search/lib/utils.js
```diff
@@ -21,8 +28,15 @@ const getIndexAlias =
...
+  (name) => [ES_INDEX_PREFIX, name, getDateTime()].filter(Boolean).join('-')
+
+/**
+ * @param  {String} name An index name without prefix
+ * @return {String}      An index name with a date
+ */
+const getDatedIndex =
```
patte: I would like `getDateIndex` better

patrickvenetz: Yes, adopting. Was under the impression, originally, that it might lead to confusion.

packages/search/lib/utils.js
patrickvenetz: No, I was not aware of that. Very good to now.

PR #81, opened 2018-08-20
WIP: feat(access): Initial version of access package to grant access

patrickvenetz:
tba


packages/access/lib/grants.js
```diff
@@ -0,0 +1,202 @@
...
+    recipientUserId: recipient.id,
+    updatedAt: moment(),
+    result
+  })
+
+  return result > 0
+}
+
+const renderInvalid = async (grant, reason, pgdb, mail) => {
+  const result = await pgdb.public.accessGrants.update(
```
lukasbuenger: Even though the name is perfectly descriptive, in our codebase I'd be hesitant to call any method renderSomething that is not actually rendering something, because that term is strongly connected to React et al.

lukasbuenger: `invalidate` would be a good alternative I think

packages/access/graphql/resolvers/AccessGrant.js
```diff
@@ -0,0 +1,40 @@
...
+
+const { Roles, transformUser } = require('@orbiting/backend-modules-auth')
+
+module.exports = {
+  campaign: (grant, args, { pgdb }) => campaignsLib.findByGrant(grant, pgdb),
+  grantee: async (grant, args, { user: me, pgdb }) => {
+    const grantee =
+      await pgdb.public.users.findOne({ id: grant.granteeUserId })
+
+    if (!Roles.userIsMeOrInRoles(grantee, me, ['admin', 'supporter'])) {
```
tpreusse: check should be before db query

patrickvenetz:
Redis lock logic tested with multiple instances with up to 15 instances.

A similar logic is in use [for publikator server's publicationScheduler](https://github.com/orbiting/backends/blob/master/servers/publikator/lib/publicationScheduler.js).

packages/access/graphql/resolvers/AccessGrant.js
```diff
@@ -0,0 +1,40 @@
...
+
+    return transformUser(grantee)
+  },
+  granteeName: async (grant, args, { user: me, pgdb }) => {
+    const grantee =
+      await pgdb.public.users.findOne({ id: grant.granteeUserId })
+
+    const safeUser = transformUser(grantee)
+
+    return safeUser.name || safeUser.email
```
tpreusse: It's currently not possible to buy or claim a membership and not have a name. I'd recommend to not fallback to the email and instead print 'Unbekannt' or 'Unbenannt'.

packages/access/graphql/resolvers/AccessGrant.js
```diff
@@ -0,0 +1,40 @@
...
+  },
+  recipient: async (grant, args, { user: me, pgdb }) => {
+    if (!grant.recipientUserId) return null
+
+    const recipient =
+      await pgdb.public.users.findOne({ id: grant.recipientUserId })
+
+    if (!Roles.userIsMeOrInRoles(recipient, me, ['admin', 'supporter'])) {
```
tpreusse: check should be before db query

packages/access/graphql/resolvers/User.js
```diff
@@ -0,0 +1,45 @@
...
+      return null
+    }
+
+    const campaigns = await campaignsLib.findForGrantee(user, pgdb)
+
+    debug('accessCampaigns', { user: user.id, campaigns: campaigns.length })
+
+    return campaigns.map(campaign => ({
+      ...campaign,
+      user: transformUser(user)
```
tpreusse: This is most likely double transforming and could lead to issues. The user here should already be transformed. To debug: `const isTransformed = !!user._raw` (don't branch in prd code just to test it locally)

packages/access/graphql/schema-types.js
```diff
@@ -0,0 +1,75 @@
...
+}
+
+"""
+Entity representing a future, current or passed granted membership
+"""
+type AccessGrant {
+  id: ID!
+  "Campaign this membership grant belongs to"
+  campaign: AccessCampaign!
+  "Entity who granted membership"
```
tpreusse: add admin only note here

packages/access/graphql/schema-types.js
```diff
@@ -0,0 +1,75 @@
...
+  grantee: User
+  "Name or email address of entity who granted membership"
+  granteeName: String!
+  """
+  Original recipient email address of grant.
+  Is eventually matched to a User (see recipient).
+  """
+  email: String!
+  "Entity who received granted membership"
```
tpreusse: admin only note

packages/access/graphql/resolvers/User.js
```diff
@@ -0,0 +1,45 @@
...
+    const users =
+      userIds.length > 0
+        ? await pgdb.public.users.find({ id: userIds })
+        : []
+
+    debug('accessGrants', { user: user.id, grants: grants.length, userIds })
+
+    return grants.map(grant => ({
+      ...grant,
+      users: users.map(transformUser)
```
tpreusse: where is this used? Is it even needed?

packages/access/graphql/resolvers/User.js
tpreusse: Where is this used? Is it even needed?

PR #82, opened 2018-08-21
Fix: Add resolved into indexed node

patrickvenetz:
This Pull Request places resolved meta information into `resolved.meta` instead of `resolved` when a document is published. Latter is not indexed and hence was not queryable.

`resolved` data servers mainly search to retrieve documents if e.g. a format title matches.

__Before, a document__

```
{
  [...]
  "resolved": {
    "format": {
      "meta": {
        "template": "format",
        "repoId": "republik/format-example",
        "description": "",
        "title": "Example Format"
      }
    }
  },
  [...]
}
```

`resolved.format` is not mapped and not indexed. This data is not queryable.

__After, a document__

```
{
  [...]
  "resolved": {
    "meta": {
      "format": {
        "meta": {
          "template": "format",
          "repoId": "republik/format-example",
          "description": "",
          "title": "Example Format"
        }
      }
    }
  },
  [...]
}
```

`resolved.meta.format` is mapped and hence indexed and queryable.

This bug affected publishing procedure but not when pulling data into ElasticSearch at once. If data is pulled into ElasticSearch, `resolved` is added properly.

It also makes sure `resolved` node in a document is only added if not empty.

### DIFF
--- a/packages/search/lib/Documents.js
+++ b/packages/search/lib/Documents.js
@@ -1,4 +1,5 @@
 const debug = require('debug')('search:lib:Documents')
+const _ = require('lodash')
 
 const {
   termEntry,
@@ -160,12 +161,12 @@ const getElasticDoc = (
     __sort: {
       date: meta.publishDate
     },
-    resolved,
     id,
     commitId,
     versionName,
     milestoneCommitId,
     meta,
+    resolved: !_.isEmpty(resolved) ? resolved : undefined,
     content: doc.content,
     contentString: mdastToString(
       mdastFilter(
--- a/servers/publikator/graphql/resolvers/_mutations/publish.js
+++ b/servers/publikator/graphql/resolvers/_mutations/publish.js
@@ -354,7 +354,8 @@ module.exports = async (
       doc.content.meta.dossier
     )
 
-    resolved.dossier = dossiers.pop()
+    if (!resolved.meta) resolved.meta = {}
+    resolved.meta.dossier = dossiers.pop()
   }
 
   if (doc.content.meta.format) {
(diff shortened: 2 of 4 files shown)


servers/publikator/graphql/resolvers/_mutations/publish.js
```diff
@@ -354,7 +354,7 @@ module.exports = async (
       doc.content.meta.dossier
     )
 
-    resolved.dossier = dossiers.pop()
+    Object.assign(resolved, { meta: { dossier: dossiers.pop() } })
```
tpreusse: `Object.assign` is not a deep merge. As far as I understand if something would belong to a format and dossier you'd loose the dossier down below

packages/search/lib/Documents.js
```diff
@@ -160,12 +161,12 @@ const getElasticDoc = (
     __sort: {
       date: meta.publishDate
     },
-    resolved,
     id,
     commitId,
     versionName,
     milestoneCommitId,
     meta,
+    resolved: !_.isEmpty(resolved) ? resolved : undefined,
```
tpreusse: Why needed? Because you now also index empty objects in the new logic? Maybe just prevent indexing empty objects in the first place.

patrickvenetz:
Code that is correctly populating `resolved` node can be found [here](https://github.com/orbiting/backends/blob/fix-resolved-node/packages/search/script/inserts/document.js#L67-L77), leading up to [`upsertResolvedMeta`](https://github.com/orbiting/backends/blob/fix-resolved-node/packages/search/script/inserts/document.js#L23-L65)

packages/search/lib/Documents.js
patrickvenetz: I do not understand this comment just yet.

PR #91, opened 2018-09-05
feat(auth): Query and submit additional fields when authorizing a session

patrickvenetz:
This Pull Request provides information on additional fields required to authorize a session not unlike `requiredConsents` field.

Query `unauthorizedSession` first:

```gql
{ unauthorizedSession(
  email: ...
  token: ...
) {
  requiredFields
} }
```

This query will result in something as follows:

```gql
{
  "data": {
    "unauthorizedSession": {
      "requiredFields": [
        "firstName",
        "lastName"
      ]
    }
  }
}
```

Authorize a session with additional required fields like so:

```gql
mutation {
  authorizeSession(
    email: ...
    tokens: ...,
    ...
    fields: {
      firstName: "Peter"
      lastName: "Parker"
    }
  )
}
```

It requires UI to updated https://github.com/orbiting/republik-frontend/pull/183

### DIFF
--- a/packages/auth/lib/Fields.js
+++ b/packages/auth/lib/Fields.js
@@ -1,3 +1,4 @@
+const debug = require('debug')('auth:lib:Fields')
 const moment = require('moment')
 
 const { newAuthError } = require('./AuthError')
@@ -8,8 +9,7 @@ const MissingFieldsError = newAuthError(
   'api/fields/missing'
 )
 
-const getMissingFields = async ({ user, email, pgdb }) => {
-  const missingFields = []
+const hasGrants = async ({ user, email, pgdb }) => {
   const unassignedGrants = await pgdb.public.accessGrants.find({
     email,
     recipientUserId: null,
@@ -18,6 +18,14 @@ const getMissingFields = async ({ user, email, pgdb }) => {
     invalidatedAt: null
   })
 
+  debug('hasGrants', unassignedGrants.length > 0)
+
+  return unassignedGrants.length > 0
+}
+
+const getMissingFields = async ({ user, email, pgdb }) => {
+  const missingFields = []
+
   const isMember = !!user && Roles.userHasRole(user, 'member')
   const hasNames = !!user && (
     user.firstName &&
@@ -25,11 +33,10 @@ const getMissingFields = async ({ user, email, pgdb }) => {
     user.lastName &&
     user.lastName.trim().length > 1
   )
-  const hasGrants = unassignedGrants.length > 0
 
   if (
     (user && isMember && !hasNames) ||
-    (!user && hasGrants)
+    (!user && await hasGrants({ user, email, pgdb }))
   ) {
     missingFields.push('firstName')
     missingFields.push('lastName')
--- a/packages/auth/lib/Fields.js
+++ b/packages/auth/lib/Fields.js
@@ -1,0 +1,61 @@
+const moment = require('moment')
+
+const { newAuthError } = require('./AuthError')
+const Roles = require('./Roles')
+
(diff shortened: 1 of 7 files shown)


packages/auth/lib/Fields.js
```diff
@@ -0,0 +1,70 @@
...
+  const unassignedGrants = await pgdb.public.accessGrants.find({
+    email,
+    recipientUserId: null,
+    'beginAt <=': moment(),
+    'endAt >': moment(),
+    invalidatedAt: null
+  })
+
+  const isMember = !!user && Roles.userHasRole(user, 'member')
+  const hasNames = !!user && (
```
tpreusse: if you do this check first you could return early without doing the grant db query if one already has a name.

packages/auth/lib/Users.js
```diff
@@ -322,13 +325,16 @@ const authorizeSession = async ({ pgdb, tokens, email: emailFromQuery, signInHoo
...
       consents,
+      fields,
       req
     }))
   } catch (error) {
     await transaction.transactionRollback()
     throw error
   }
 
+  // throw new Error('Breakpoint')
```
tpreusse: 🚿

packages/auth/lib/Users.js
```diff
@@ -406,18 +412,23 @@ const upsertUserAndConsents = async ({ pgdb, email, consents, req }) => {
...
+    ? await pgdb.public.users.insertAndGet({
+      ...userFields,
       email,
       verified: true
     })
-  if (!user.verified) {
-    await pgdb.public.users.updateOne({
-      id: user.id
-    }, {
+    : await pgdb.public.users.updateAndGetOne({ id: existingUser.id }, {
```
tpreusse: I'd avoid always updating the user object on every `authorizeSession`. Check if userFields actually contains something. Also missing `updatedAt: new Date()`.

packages/auth/lib/Fields.js
```diff
@@ -0,0 +1,70 @@
...
+  const expectedFields = await getMissingFields({ user, email, pgdb })
+
+  expectedFields
+    .forEach(expectedField => {
+      if (!providedFields[expectedField]) {
+        missingFields.push(expectedField)
+        return false
+      }
+
+      fields[expectedField] = providedFields[expectedField]
```
tpreusse: IMO this is a bad approach: if one sends fields but they are not missing on the user they are ignored without a warning.

You don't have to worry about users sending additional fields through the mutation resolver:
```gql
mutation {
  authorizeSession(
    email: "test",
    tokens: [{type: EMAIL_TOKEN, payload: "lala"}],
    fields: {
      firstName: "Troll",
      lastName: "Leet",
      birthday: "dead"
    }
  )
}
```
will result in
```json
{
  "errors": [
    {
      "message": "Field \"birthday\" is not defined by type RequiredUserFields.",
      "locations": [
        {
          "line": 8,
          "column": 7
        }
      ]
    }
  ]
}
```

patrickvenetz: I understand.

Did not worry about additional fields, fully aware GraphQL schema would return an error if undefined.

Thinking behind this flow was this:

You can only update fields what `unauthorizedSession` lists in `requiredFields`. This would prevent the ability to update first and last name "unwarranted" during `authorizeSession`.

I still believe that is a healthy, safe way. We could throw an error if there are not required fields provided. This or process whatever is provided.

Not feeling strong about it.

tpreusse: Ok but then call the argument `requiredFields` in the `authorizeSession` mutation, that way there is a clear hint that it will only do when previously required.

PR #94, opened 2018-09-11
feat(security): wear helmet

tpreusse:
Introduce all easy to implement security headers (skip CSP and HPKP for now).

HSTS initially just with a 24h window to test and ensure that no subdomain services breaks.

### DIFF
--- a/packages/base/server.js
+++ b/packages/base/server.js
@@ -132,7 +132,9 @@ const start = async (
       // preload: true,
       includeSubDomains: true
     },
-    referrerPolicy: true
+    referrerPolicy: {
+      policy: 'no-referrer'
+    }
   }))
 
   // prod only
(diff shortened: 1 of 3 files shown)


packages/base/server.js
```diff
@@ -124,6 +125,16 @@ const start = async (
...
   httpServer = createServer(server)
 
+  server.use(helmet({
+    hsts: {
+      maxAge: 60 * 60 * 24, // one day for initial testing
+      // maxAge: 31536000, // 1 year to get preload approval
+      // preload: true,
+      includeSubDomains: true
+    },
+    referrerPolicy: true
```
patrickvenetz: For the sake of clarity would set explicit (albeit [default](https://helmetjs.github.io/docs/referrer-policy/)) value: `no-referrer`

patrickvenetz:
LGTM

Since HSTS is not pinning (HPKP that is, my bad) but just signaling a browser to request over HTTPS, I would feel confident to increase that one. As far as I can tell we do not deliver any insecure resources.

tpreusse:
Thanks for the feedback. Made the referrer policy explicit. I want one day of HSTS initially because it includes all subdomains and if we forgot about something insecure we could revert within 24h.

PR #99, opened 2018-09-25
Fix republik deleteUser mutation

patrickvenetz:
This Pull Request fixes `deleteUser` mutation.

It keeps nulled `user` entity if not only there is a pledge [but access grants were issued or received](https://github.com/orbiting/backends/compare/fix-delete-user-query?expand=1#diff-6f5a7eacb382d487b08306d78b931f97R147).

Records in `accessGrants` are kept as is. Deleting would create sync issues with regular revoke/invalidate procedure. Attribute `email` is kept as is as it is originally provided by grantee and not recipient (it is exposed to grantee).

It then fixes notification listener in `search` package to be [more](https://github.com/orbiting/backends/compare/fix-delete-user-query?expand=1#diff-4bd3489e5cea952d1b51961d49681c5aR39) [lenient](https://github.com/orbiting/backends/compare/fix-delete-user-query?expand=1#diff-791481b469dfa11922492ed8de641bc4R65) on data that no longer exists. `deleteUser` mutation will delete data before notification listener can fetch user (and it's belongings) and post to ElasticSearch.

A foreign key attribute no longer requires [to be named `userId`](https://github.com/orbiting/backends/compare/fix-delete-user-query?expand=1#diff-6f5a7eacb382d487b08306d78b931f97R79).

TODO
- [x] Should not delete records in comments but anonymize them, and maybe clear content. Deleting a comment would potentially destroy trees and leave unwanted "loose" comments.
- [ ] make PROLONG pledges cancellable otherwise all users who renewed cannot be deleted.

deploy:
- [ ] migrate DB


patrickvenetz:
⚠️ Should not delete records in `comments` but anonymize them, and maybe clear content. Deleting a comment would potentially destroy trees and leave unwanted "loose" comments.

PR #107, opened 2018-10-02
feat(access): Send expiration and follow up transactional email

patrickvenetz:
This Pull Request adds missing pieces to granting access feature:
- send a transactional email at the moment a grant expires
- send a transactional email to follow-up some time later (set in campaign itself)

Other changes include:
- Handle errors in `accessScheduler` more gracefully. Mandrill API proves to be not a reliable and throws frequently 500 HTTP Status Code errors containing not parseable content.
- Includes more verbose debug logging to understand who queried what and when.
- Changed schedulers interval to 10 minutes as matching happens upon sign in.

### DIFF
--- a/packages/access/lib/accessScheduler.js
+++ b/packages/access/lib/accessScheduler.js
@@ -105,7 +105,7 @@ const matchGrants = async (pgdb, mail) => {
  * Renders expired grants invalid.
  */
 const expireGrants = async (t, pgdb, mail) => {
-  debug('expireGrants')
+  debug('expireGrants...')
   for (const grant of await grantsLib.findExpired(pgdb)) {
     const transaction = await pgdb.transactionBegin()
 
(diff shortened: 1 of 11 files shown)


packages/access/lib/accessScheduler.js
```diff
@@ -29,7 +30,8 @@ const init = async ({ mail }) => {
         await schedulerLock()
           .lock('locks:access-scheduler', 1000 * intervalSecs)
```
patte: should the TTL not decrease here? Am I right that you lock here for all of the interval, just to be sure?

packages/access/lib/grants.js
```diff
@@ -183,13 +195,56 @@ const invalidate = async (grant, reason, pgdb, mail) => {
...
+        grantee, campaign, recipient, grant, t, pgdb
+      )
+    }
+  }
+
+  debug('followUp', {
+    id: grant.id,
+    hasRecipient: !!grant.recipientUserId,
+    followUp: moment(),
+    updatedAt: moment(),
```
patte: between this debug and the db call on line 199 potentially 4 async calls, why not use the real dates here?

packages/access/lib/accessScheduler.js
patrickvenetz: Not sure why it should be decreased.

It does indeed lock the entire interval. If it succeeds, it will [extend lock by a 10th](https://github.com/orbiting/backends/pull/107/files/0488617e44f33bf09ba9506d8f6301cbc100b92f#diff-41d11ac02aa1af5acae07bf411f9bfccR38) to prevent back-to-back runs (and somewhat decrease it).

packages/access/lib/grants.js
patrickvenetz: Will do.

PR #109, opened 2018-10-03
feat(preview): Send followup mail 7 days after a preview request expired

patrickvenetz:
This Pull Request extends preview scheduler to check for expired preview requests and send a follow-up email if request expired more than 7 days ago. Only users with no active membership will receive an email.

It is ought to replace MailChimp campaigns (and lists) which were created and mailed each month to preview recipients.

(Before releasing, at least _preview_followup_ templates requires to be updated.)


packages/preview/lib/mail.js
```diff
@@ -55,7 +99,13 @@ const getGlobalMergeVars = () => ([
...
     content: `${FRONTEND_BASE_URL}/impressum`
   },
+  { name: 'LINK_IMPRESSUM',
+    content: `${FRONTEND_BASE_URL}/impressum`
+  },
   { name: 'LINK_PROJECTR',
     content: 'https://project-r.construction/'
+  },
+  { name: 'LINK_PROJECTR_NEWS',
+    content: 'https://project-r.construction/news'
```
patte: These urls should all go to translations, as they are heavily specific to our environment.

packages/preview/lib/preview.js
```diff
@@ -104,6 +105,42 @@ const expire = async (requests, users, pgdb, mail) => {
...
+    requests,
+    async (request) => {
+      const transaction = await pgdb.transactionBegin()
+
+      try {
+        const user = users.find(user => user.id === request.userId)
+        const hasMembership =
+          !!memberships.find(membership => membership.userId === user.id)
+
+        await setFollowup(request, user, pgdb)
```
patte:
```suggestion
        await setFollowup(request, user, transaction)
```

packages/preview/lib/preview.js
```diff
@@ -124,6 +161,17 @@ const findVoidable = (pgdb) => {
   })
 }
 
+const findEmptyFollowup = (pgdb) => {
```
patte: returns a promise, so it should be async
```suggestion
const findEmptyFollowup = async (pgdb) => {
```

packages/preview/lib/preview.js
```diff
@@ -104,6 +105,42 @@ const expire = async (requests, users, pgdb, mail) => {
...
+      } catch (e) {
+        await transaction.transactionRollback()
+
+        debug('rollback', { request: request.id })
+
+        throw e
+      }
+    },
+    { concurrency: 5 }
+  )
```
patte: Can one user have multiple open requests?

patrickvenetz: A user can have multiple subsequent requests. A user can resubmit a request when other request is expired.

packages/preview/lib/preview.js
patrickvenetz: Interesting. Thus far I was under the impression I'd have to declare a function `async` only if I do async/await or resolve/wait for Promises inside a function. Here I basically only "declare" a promise that someone else will have to resolve.

Let me check on that.

packages/preview/lib/mail.js
patrickvenetz: Not sure how I can manage to make it agnostic. Any hints here?

packages/preview/lib/preview.js
patte: Ok, so this is not going to send multiple emails to one user, because each user can only have one request where `expireAt` is null?

patrickvenetz: Yes.

Starts [with `begin` function](https://github.com/orbiting/backends/blob/da8bec239f0093ddbe243dba59f99c8773dfca72/packages/preview/lib/preview.js#L38-L42) which queries [for "other" (valid) request](https://github.com/orbiting/backends/blob/da8bec239f0093ddbe243dba59f99c8773dfca72/packages/preview/lib/preview.js#L175-L180) before adding request.

packages/preview/lib/preview.js
patrickvenetz: There is no particular example available, but [this may be something along the lines](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/async_function#Simple_example). `findEmptyFollowup` returns an unresolved Promise on purpose.

PR #112, opened 2018-10-05
fix(series): sanitize repo ids before querying

tpreusse:
Following values are in `meta.series`:
```
https://github.com/republik-dev/article-daniels-serienmaster
https://github.com/republik-dev/article-master
https://github.com/republik-dev/article-master
https://github.com/republik/article-intro
```
after passing through `getRepoId` `"https://github.com/republik/article-intro"` becomes `null` because it's not a recognized id (different org).

Querying for "meta.repoId: ["...", null]" leads to an elastic parse error:

```
[parsing_exception] No value specified for terms query, with { line=1 & col=199 }
```

### DIFF
--- a/packages/search/lib/Documents.js
+++ b/packages/search/lib/Documents.js
@@ -271,14 +271,15 @@ const addRelatedDocs = async ({ connection, scheduledAt, context }) => {
 
   // If there are any series master repositories, fetch these series master
   // documents and push series episodes onto the related docs stack
-  if (seriesRepoIds.length > 0) {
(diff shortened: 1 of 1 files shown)


patte:
In your case, adding `republik` to the `GITHUB_ORGS` env var, should correctly parse the url (see [resolve.js](https://github.com/orbiting/backends/blob/master/packages/documents/lib/resolve.js#L37)). I still think we should merge this, as we sanitize repoIds (for related docs) the same way.

PR #117, opened 2018-10-15
keepalive middleware for /graphql

patte:
enable with RES_KEEPALIVE=true env
sends (headers and) whitespace until response is ready

used to trick into heroku to think a response is being sent and thus avoids that the request gets rescheduled to other workers

source: https://spin.atomicobject.com/2018/05/15/extending-heroku-timeout-node/

### DIFF
--- a/packages/base/express/keepalive.js
+++ b/packages/base/express/keepalive.js
@@ -1,0 +1,53 @@
+// https://spin.atomicobject.com/2018/05/15/extending-heroku-timeout-node/
+const debug = require('debug')('base:keepalive')
+
+const {
+  RES_KEEPALIVE_MS = 15000
+} = process.env
+
+module.exports = (req, res, next) => {
+  const space = ' '
+  let isFinished = false
+  let isDataSent = false
+
+  res.once('finish', () => {
+    isFinished = true
+  })
+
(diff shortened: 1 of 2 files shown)


packages/base/express/keepalive.js
```diff
@@ -0,0 +1,53 @@
+// https://spin.atomicobject.com/2018/05/15/extending-heroku-timeout-node/
+const debug = require('debug')('base:keepalive')
+
+const {
+  RES_KEEPALIVE_MS = 15000
```
patrickvenetz:
```suggestion
  RES_KEEPALIVE_MS = 1000 * 15
```

packages/base/express/keepalive.js
```diff
@@ -0,0 +1,53 @@
...
+      if (!isFinished && !isDataSent) {
+        // Need to write the status code/headers if they haven't been sent yet.
+        if (!res.headersSent) {
+          res.writeHead(202)
+        }
+
+        res.write(space)
+        debug('sent keepalive')
+
+        // Wait another TIMEOUT_MS
```
patrickvenetz:
```suggestion
        // Wait another RES_KEEPALIVE_MS
```

PR #121, opened 2018-10-17
feat(base): rate limiter

patte:
- rate limiter for the base modul (used by republik and publikator backend)
- send a slack message if a connection is blocked (on first occasion)

ENVs:
- set `RATE_LIMIT_DISABLE=true` to disable it
- set `X_SSR_FOR_HMAC_KEY` to the same value the frontend uses

accompanying frontend PR: https://github.com/orbiting/republik-frontend/pull/210

parse logs for `ratelimit`
```
ratelimit reached for user: `u1054752-eefe-4cb4-9da0-b5aa9c07d334` IP: `::1` (key: `b3054752-eefe-4cb4-9da0-b57a9c07d334_::1`)
```

### deploy:
- [ ] set `X_SSR_FOR_HMAC_KEY` in BE and FE
- [ ] set `RATE_LIMIT_DISABLE` in publikator-api and publikator-api-staging
- [ ] limit google bot to 2rps


tpreusse:
```js
const express = require('express')
const app = express()

const PORT = process.env.PORT || 3333

app.enable('trust proxy')
app.get('/', (req, res) => {
  res.send([
    `IP: ${req.ip}`,
    `IPs: ${req.ips.join(', ')}`,
    ''
  ].join('\n'))
})

app.listen(PORT, () => console.log(`Listening on port ${PORT}!`))
```

We can't use `req.ip` 😬 

```
00:45 ~ $ curl -H 'X-Forwarded-For: 8.8.8.8' https://ip-fwd-test.herokuapp.com/
IP: 8.8.8.8
IPs: 8.8.8.8, 212.51.140.102
```

patte:
You gotta be kidding me!!! Why does heroku not have a basic rate limiter, why can't we block IPs, why is the timeout not configurable?? Anyway, thanks for making the test @tpreusse. I'll adapt.

patte:
If have to unrant (a little bit), it acutally makes sense, that proxies add the connecting IP to the right of `X-FORWARDED-FOR`. And cloudflare also does it that way. And still we should use `req.ip`, but also fix the configuration: you and the backends used `app.enable('trust proxy')` this just trusts all proxies (also the client if he pretends to be one). It's crucial to set how many proxies to trust and use `app.set` for it, so: `app.set('trust proxy', 1)` in our case. [see](http://expressjs.com/en/guide/behind-proxies.html)

## case 1: Let's say we add a proxy in front of heroku. 

client has IP: 1.2.3.4
client sends malicious `X-FORWARDED-FOR=8.8.8.8`
after first proxy: `X-FORWARDED-FOR=[8.8.8.8,1.2.3.4]`
after second proxy (heroku): `X-FORWARDED-FOR=[8.8.8.8,1.2.3.4,1.1.1.1]`

with `app.set('trust proxy', 2)` we get `req.ip = 1.2.3.4` which is correct

## case 2: only one proxy

client has IP: 1.2.3.4
client sends malicious `X-FORWARDED-FOR=8.8.8.8`
after first proxy: `X-FORWARDED-FOR=[8.8.8.8,1.2.3.4]`

with `app.set('trust proxy', 1)` we get `req.ip = 1.2.3.4` which is correct

## case 3: non malicious client
client has IP: 1.2.3.4
after first proxy: `X-FORWARDED-FOR=[1.2.3.4]`

with `app.set('trust proxy', 1)` we get `req.ip = 1.2.3.4` which is correct

I will adapt accordingly. It's also something to keep in mind if we ever happen to add another proxy infront.

PR #123, opened 2018-10-19
fix(documents): Retrieve credits in a more appropriate manner

patrickvenetz:
This Pull Request rewrites changes made in https://github.com/orbiting/backends/pull/103. I only came hereafter to understand concept of `doc.meta.credits` and `doc.content` and credit nodes.

There are two paths retrieving credits for documents:

- via `doc.content` and retrieving (visiting) meta children (Publikator and Search, default path) 
- Search may sometimes not provide (`doc.content`) due to performance considerations. In that case we attempt to retrieve credits via `doc.meta` (data stored in ElasticSearch).

Latter is currently the case when using search in frontend.

### DIFF
--- a/packages/documents/lib/meta.js
+++ b/packages/documents/lib/meta.js
@@ -3,42 +3,42 @@ const visit = require('unist-util-visit')
 const { metaFieldResolver } = require('./resolve')
 
 /**
- * If not available already, assign credits to doc.content.meta.
+ * Obtain credits from either {doc.content.children} or {doc.meta}.
  *
  * @param  {Object} doc An MDAST tree
- * @return {Object}     Returns maybe altered {doc}
+ * @return {Array}      MDAST children
  */
-const ensureCredits = doc => {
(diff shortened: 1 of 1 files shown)


packages/documents/lib/meta.js
```diff
@@ -3,53 +3,78 @@ const visit = require('unist-util-visit')
...
+  return (doc.meta && doc.meta.credits) || []
 }
 
+/**
+ * Builds and an audioSource object from {doc.content.meta} for use in meta.
+ *
+ * @param  {Object}      doc An MDAST tree
+ * @return {Object|null}     e.g. { mp3: true, aac: null, ogg: null }
+ */
+const getAudioSource = doc => {
```
patte:
```suggestion
const getAudioSource = doc => {
  if (!doc.content && doc.meta && doc.meta.audioSource) {
    return doc.meta.audioSource
  }
```
This might help for the upcoming feature of showing an audio icon in the feed/search result.

PR #125, opened 2018-10-23
fix(Discussion): default false value for collapsable

uxengine:


servers/publikator/lib/Discussion.js
```diff
@@ -6,7 +6,7 @@ const upsert = async (repoMeta, docMeta, context) => {
   const settings = {
     title: docMeta.title,
     documentPath: docMeta.path,
-    collapsable: docMeta.collapsable,
+    collapsable: docMeta.collapsable !== undefined ? docMeta.collapsable : false,
```
patrickvenetz:
```suggestion
    collapsable: !!docMeta.collapsable,
```

PR #126, opened 2018-10-24
fix(crowdfundings): Alter conditions when to accept a voucher code

patrickvenetz:
This Pull Request extends checks who and when a voucher code can be claimed.

Conditions:
1.  _Now:_ Checks if user has already an active membership and throws an error if so.
    (_Before:_ Checked if user had a membership at all and throws and error if so.)
2.  _Now:_ Checks if user owns voucher code and throws and error if so.
    (_Before:_ Check was not necessary, as previous check did decline voucher code claim attempt already.)

[Querying for a valid voucher code](https://github.com/orbiting/backends/compare/fix-claim-memberships?expand=1#diff-7c64fc3b0eae9067f7921e879575d28bR19) becomes more strict, based on [later membership update statement](https://github.com/orbiting/backends/compare/fix-claim-memberships?expand=1#diff-7c64fc3b0eae9067f7921e879575d28bR45)

As of today, a user can not claim their own membership.


servers/republik/modules/crowdfundings/graphql/resolvers/_mutations/claimMembership.js
```diff
@@ -5,23 +5,37 @@ const moment = require('moment')
 module.exports = async (_, args, {pgdb, req, t, mail: {enforceSubscriptions}}) => {
   ensureSignedIn(req)
 
-  // if this restriction gets removed, make sure to check if
-  // the membership doesn't already belong to the user, before
-  // making the the transfer and removing the voucherCode
-  if (await pgdb.public.memberships.count({userId: req.user.id})) {
+  // Throw an error if there is already an other active membership.
```
patte: Reading it now, I think everything from line 8 to 32 should also go into the transaction to the consistency of the state between querying and modifying.

PR #127, opened 2018-10-24
fix(Discussion): comment preview for unpublished comments

tpreusse:
According to the schema preview can be null but `Preview.string` not:
https://github.com/orbiting/backends/blob/ad8170c7899a071f7da6cb9af15b9cdc781dea0c/servers/republik/graphql/schema-types.js#L304-L307

https://github.com/orbiting/backends/blob/ad8170c7899a071f7da6cb9af15b9cdc781dea0c/servers/republik/graphql/schema-types.js#L323-L326

The current implementation explodes with unpublished text with `Cannot return null for non-nullable field Preview.string`.

However it looks like `preview` might be called somewhere else, for notifications?, too but I couldn't find it to verify if that can handle null instead of `{text: null: more: false}` (which is not gql schema compliant). Alt would be to return `{text: '', more: false}`.

### DIFF
--- a/servers/republik/graphql/resolvers/Comment.js
+++ b/servers/republik/graphql/resolvers/Comment.js
@@ -88,10 +88,7 @@ module.exports = {
   preview: (comment, { length = 500 }, context) => {
     const text = textForComment(comment, context && context.user)
     if (!text) {
-      return {
-        string: text,
-        more: false
-      }
+      return null
     }
     return mdastToHumanString(remark.parse(text), length)
   },


patrickvenetz:
Body for notification is a comment string (with a prepended author display name):

https://github.com/orbiting/backends/blob/070e488c7f45869605206ee08f2844680d46c523/servers/republik/lib/Notifications.js#L130-L138

`shortBody` is [a shortened string of `comment`](https://github.com/orbiting/backends/blob/070e488c7f45869605206ee08f2844680d46c523/servers/republik/lib/Notifications.js#L92-L94). (Code is not considering that `comment` may contain Markdown.)

patrickvenetz:
@tpreusse Can you point me towards prod env where it explodes when text is unpublished?

uxengine:
FYI, we noticed yesterday in my pending https://github.com/orbiting/republik-frontend/pull/200 when I retrieved `preview` for all comments of a discussion. I've now changed things to only retrieve `preview` for the focus comment, but we should still fix it.

It also doesn't currently explode in search, because unpublished comments are excluded, but it may explode in the future if someone does things like I did.

PR #132, opened 2018-11-02
Extensions

patte:
- [ ] SQL statement [to create crowdfunding phase, package and options](https://gist.github.com/patrickvenetz/316676cb7ac0407688b66b82d86f0322).
- [ ] `cat servers/republik/local/cancellations.csv | heroku run 'node servers/republik/script/prolong/importCancellations.js' -a republik-api`
- [ ] `heroku run 'cd servers/republik && DEBUG=null node script/prolong/updateMailchimp.js --dry' -a republik-api`


patrickvenetz:
I believe it won't be in issue at the moment.

While a pledge is based on a package that distinguishes which crowdfunding phase there was, there should be little harm to having a membership extension bought from another crowdfunding-package-phase.

patrickvenetz:
Came to settle for this structure:

```gql
{ 
  me {
    customPackages {
      id
      name
      options {
        id
        templateId
        reward {
          ... on MembershipType {
            name
            interval
            intervalCount
          }
        }
        customization {
          membershipId
          bonusInterval
          bonusIntervalCount
        }
      }
    }
  }
}

```

Deviations from what we agreed on:

- Custom package props are to be found in `me.customPackages.customization`, adding (sane?) structure and allowing for other extensions
- `me.customPackages.customization.bonusInterval[Count]` props instead of `bonusDays`, adopting count/interval pattern from in `me.customPackages.rewards.interval[Count]` (for `MembershipTypes). Allows a less stiff bonus intervals.

Resulting in a response similar to:

```json
{
  "data": {
    "me": {
      "customPackages": [
        {
          "id": "00000000-0000-0000-0000-000000000000",
          "name": "PROLONG_ABO",
          "options": [
            {
              "id": "00000000-0000-0000-0000-000000000000-00000000-0000-0000-0000-000000000000",
              "templateId": "00000000-0000-0000-0000-000000000000",
              "reward": {
                "name": "ABO",
                "interval": "year",
                "intervalCount": 1
              },
              "customization": {
                "membershipId": "00000000-0000-0000-0000-000000000000",
                "bonusInterval": "days",
                "bonusIntervalCount": 1
              }
            },
            ...
          ]
        }
      ]
    }
  }
}
```

WIP-extension.md
```diff
@@ -1,7 +1,33 @@
...
+Input:
+- A membership
+- A prolong packageOption, to check against
+
+a)  Does user own membership, or pledge membership (ABO_GIVE)
+    - If false, end and don't return package option
+b)  Is membership active
+    - If false, end
+c)  Is current membershipPeriod last period
+    - If false, end
```
patte: c is optional I think. It's not mandatory to avoid extensions of extensions.

WIP-extension.md
```diff
@@ -1,7 +1,33 @@
...
+d) Can membership.type be extend
+    - If false, end
+e)  If membership.membershipType != option.membershipType
+    - If true, indicate generation of new membership in customization
+    payload, then proceed
```
patte: This is for upgrades/downgrades right? If so option should have something like compatibleMembershipTypes and this check would be `if option.compatibleMembershipTypes.indexOf(membership.membershipType) > -1`

WIP-extension.md
```diff
@@ -1,7 +1,33 @@
...
+Flow to determine whether a prolong `packageOption` is returned.
+
+Input:
+- A membership
+- A prolong packageOption, to check against
+
+a)  Does user own membership, or pledge membership (ABO_GIVE)
+    - If false, end and don't return package option
+b)  Is membership active
+    - If false, end
```
patte: This would mean that memberships that have ended because they were not extended, need to be replaced by a new membership right? Should we not make it possible to reactive by extending?

patrickvenetz: Thus far in my mind a `membership` should have `membershipPeriods` without a gap in between.

But the longer I think about it, it appears not to be a necessity.

Exploring potential downsides: Having gaps however would come at the expense of potentially more complex queries in database to retrieve statistical, financial data. If we were to allow `active: false` memberships to be extended, we'd probably need another check to prevent unclaimed "gift" memberships to be extended to eternity.

Sounds to me like a business decision.

WIP-extension.md
patrickvenetz: It is, yes.

There is currently no scenario planned on a business in which you could not switch from one type to the other e.g. you can't switch from `BENEFACTOR` to `ABO_GIVE`-esque membership.

WIP-extension.md
patte: > Exploring potential downsides: Having gaps however would come at the expense of potentially more complex queries in database to retrieve statistical, financial data.
this is true. We need to do it anyways as we are already today extending (reactivateMembership) monthly memberships that have been deactivated before. The logic there is that everybody has max. 1 monthly memberhip. If it's canceld it's dormant (active=false) until it get's reactivated (new period, active=true)

> we'd probably need another check to prevent unclaimed "gift" memberships to be extended to eternity.
That should not be a problem as they are not active and don't have a membershipPeriod yet.

WIP-extension.md
patte: > There is currently no scenario planned on a business in which you could not switch from one type to the other e.g. you can't switch from BENEFACTOR to ABO_GIVE-esque membership.

Can you elaborate? I read this two statements as oposing each other.
As you cannot switch from any membershipType to any other I would store in the membershipTypes table to which membershipTypes one can be upgraded. Or vice-versa: a membershipType sais from which membershipType can be upgraded to it.

patrickvenetz: There is no (business) scenario planned in which you can switch from `BENEFACTOR` to `ABO_GIVE`. A benefactor membership would allow to extend with `ABO` or `BENEFACTOR`. But you can't turn your benefactor membership into a "giftable" membership.

`customPackages` won't provide you with a `ABO_GIVE` "package option".

You can extend a membership you pledged but gave away of `membershipType` `ABO` (there is no `ABO_GIVE` type). That transition is typoe `ABO` to `ABO`.

Am I making any sense? Am I missing here something?

WIP-extension.md
patrickvenetz: Feels like this would have implications on finance. Added to question list.

patrickvenetz:
Updated structure in `user.customPackages` (or `me.customPackages`):

```gql
query getUser($slug: String) {
  user(slug: $slug) {
    email
    customPackages {
      id
      name
      options {
        id
        templateId
        reward {
          ... on MembershipType {
            name
          }
        }
        customization {
          membership {
            id
            type {
              name
            }
            periods {
              kind
              beginDate
              endDate
            }
          }
          additionalPeriods {
            kind
            beginDate
            endDate
          }
        }
      }
    }
  }
}
```

Contains `customPackages` node, containing potential custom packages. That is, as of now, only `PROLONG`.

Prolong package contains `options` with available memberships to extend. A membership may appear one than once. It differs in `options.reward` (e.g. membership A as `ABO`, membership A as `BENEFACTOR_ABO`, etc.).

```json
{
  "customPackages": [
    {
      "id": "00000000-0000-0000-0000-000000000000",
      "name": "PROLONG",
      "options": [
        {
          "templateId": "00000000-0000-0000-0000-000000000000",
          "reward": {
            "name": "BENEFACTOR_ABO"
          },
          "customization": {
            "membership": {
              "id": "00000000-0000-0000-0000-000000000000",
              "type": {
                "name": "ABO"
              }
            }
          }
        },
        {
          "templateId": "00000000-0000-0000-0000-000000000000",
          "reward": {
            "name": "ABO"
          },
          "customization": {
            "membership": {
              "id": "00000000-0000-0000-0000-000000000000",
              "type": {
                "name": "ABO"
              }
            }
          }
        }
      ]
    }
  ]
}
```

Membership it refers to is returned in `options.customization.membership` and exposes calculated, foreseen `options.customization.additionalPeriods`.

```json
{
  "additionalPeriods": [
    {
      "kind": "REGULAR",
      "beginDate": "2019-05-14T18:44:37.968Z",
      "endDate": "2020-05-14T18:44:37.968Z"
    },
    {
      "kind": "BONUS",
      "beginDate": "2020-05-14T18:44:37.968Z",
      "endDate": "2020-07-18T08:52:13.387Z"
    }
  ]
}
```

Periods are `MembershipPeriod` entities as they would be created if pledge is submitted and payed. Same piece of code creating those additional periods determines upon generating membership what periods are actually added to database.

`MembershipPeriod` received a "classification" with `kind` prop and helps us to distinguish regular periods from bonus periods. Regular periods are generated via a defined membership interval (as is) and bonus periods are generated due to other factors.

patrickvenetz:
Meanwhile: How to setup locally is as easy as shoving Buittoni Piccolinis into the stove:

1.  Checkout this branch
2.  Install dependencies

        yarn

3.  Fetch a fresh, new database dump
  
        heroku pg:pull (...)

4.  Update `.env` in backends root directory for more convenient development:
4.1 Set `DATABASE_URL` in `.env` if you pulled into a new database
4.2 Disabled sending emailss: `SEND_EMAILS=false`
4.3 Disable access grants scheduler: `ACCESS_SCHEDULER_OFF=true`
4.4 Disable preview requests scheduler: `PREVIEW_SCHEDULER_OFF=true`
4.5 Disable publication scheduler (optional): `PUBLICATION_SCHEDULER_OFF=true`

5.  Run migrations

        yarn db:migrate:up

6.  Run SQL statements from [WIP-extension.sql](https://github.com/orbiting/backends/blob/40edfa09cce0a1637012ae8dac61463b7149de0d/WIP-extension.sql)
7.  Develop

WIP-extension.sql
```diff
@@ -0,0 +1,67 @@
...
+    "price",
+    "userPrice",
+    "minUserPrice",
+    "vat"
+  )
+VALUES
+  (
+    E'6b8897bf-7ab4-433c-92a1-64fafcd54417',
+    E'3837d7e2-8b8e-4e9a-8106-4de5aacedbcd',
+    1,
```
tpreusse: `minAmount` means that one has to purchase at least one to complete the pledge. Should be 0.

WIP-extension.sql
```diff
@@ -0,0 +1,67 @@
...
+    24000,
+    FALSE,
+    0,
+    0
+  ),
+  (
+    E'6b8897bf-7ab4-433c-92a1-64fafcd54417',
+    E'5d4ab411-8a26-40dd-8a4f-f59cbfc9503f',
+    1,
```
tpreusse: same here

WIP-extension.sql
```diff
@@ -0,0 +1,67 @@
...
+  )
+VALUES
+  (
+    E'6b8897bf-7ab4-433c-92a1-64fafcd54417',
+    E'3837d7e2-8b8e-4e9a-8106-4de5aacedbcd',
+    1,
+    1,
+    1,
+    24000,
+    FALSE,
```
tpreusse: `userPrice` should probably be true.

servers/republik/modules/crowdfundings/graphql/schema-types.js
```diff
@@ -63,13 +67,27 @@ type PackageOption {
   createdAt: DateTime!
   updatedAt: DateTime!
 
+  # for pledgeOptions
   amount: Int
-  templateId: ID
+
+  # for custom packages
+  customization: PackageOptionCustomization
```
tpreusse: I'd vote for flattening: `membership` and `membershipPeriods` directly on the option.

servers/republik/modules/crowdfundings/graphql/schema-types.js
```diff
@@ -63,13 +67,27 @@ type PackageOption {
...
+}
+
 input PackageOptionInput {
   amount: Int!
   price: Int!
-  templateId: ID!
+  templateId: ID! # packageOption.id
+
+  # via custom packages
+  customization: PackageOptionCustomizationInput
```
tpreusse: same here. IMO flatten: `membershipId` directly on `PackageOptionInput`

servers/republik/modules/crowdfundings/graphql/resolvers/User.js
```diff
@@ -75,6 +77,73 @@ module.exports = {
...
+          membershipId: memberships.map(membership => membership.id)
+        })
+        : []
+
+    memberships.forEach((membership, index, memberships) => {
+      memberships[index].membershipPeriods =
+        membershipPeriods.filter(membershipPeriod => membershipPeriod.membershipId === membership.id)
+    })
+
+    Object.assign(user, { memberships })
```
patte: why do you mutate user here? you are in a recursive resolver flow, downstream resolvers might get confused.

tpreusse:
`autoPay` should be available on `pledgeOption`, and default to `true` for options selling new memberships and renewing ones own membership (`option.membership.user.id === (me && me.id)`). Return `false` on `pledge.options.autoPay` (draft pledge) query if previously submitted with false. And `null` if not applicable. We need to re-access that on `ReceivePayment` to preserve the checkbox state when one return from a PSP (e.g. 3-d secure). This will also allow to reconstruct with which pledge one turned on `autoPay` via api, can potentially be displayed in admin tool.

WIP-extension.sql
```diff
@@ -0,0 +1,69 @@
...
+  )
+VALUES
+  (
+    E'6b8897bf-7ab4-433c-92a1-64fafcd54417',
+    E'3837d7e2-8b8e-4e9a-8106-4de5aacedbcd',
+    0,
+    1,
+    0,
+    24000,
+    TRUE,
```
tpreusse: re `userPrice` true here, it's a bit more complicated: it should be true for ones own abo and false for abo give renewals. Needs some code logic.

servers/republik/modules/crowdfundings/lib/CustomPackages/rules/earlyAdopterBonus.js
```diff
@@ -0,0 +1,57 @@
+const debug = require('debug')('crowdfundings:lib:CustomPackages:rule:earlyUserBonus')
+const moment = require('moment')
+const uuid = require('uuid/v4')
+
+const MEMBERSHIP_CREATED_BEFORE = '2017-04-27'
+
+module.exports = ({ package_, packageOption, membership, payload, now }) => {
+  if (membership.createdAt >= moment(MEMBERSHIP_CREATED_BEFORE)) {
```
tpreusse: ideally the rule would be
```
membership.pledge.createdAt && membership.pledge.package.name !== 'ABO_GIVE'
```

currently memberships from `ABO_GIVE` gen on 26. would get it too right?

patrickvenetz: > currently memberships from ABO_GIVE gen on 26. would get it too right?

Yes.

Excluding `ABO_GIVE` memberships should be fairly straight forward.

WIP-extension.sql
patrickvenetz: Understood.

servers/republik/modules/crowdfundings/lib/CustomPackages/rules/earlyAdopterBonus.js
patrickvenetz: Adopted in 13a5e29

WIP-extension.sql
patrickvenetz: Adopted @ 9002825

servers/republik/modules/crowdfundings/lib/generateMemberships.js
```diff
@@ -60,52 +62,123 @@ module.exports = async (pledgeId, pgdb, t, req, logger = console) => {
...
+            .map(period => omit(period, ['id', 'createdAt', 'updatedAt']))
+        )
+
+        await pgdb.public.memberships.update(
+          { id: membershipId },
+          {
+            autoPay,
+            active: true,
+            renew: true,
+            updatedAt: now
```
patte: why is the `membershipTypeId` not updated here. If I upgrade on prolong I want to have the approriate type from then on, or how is this achived?

servers/republik/modules/crowdfundings/graphql/schema-types.js
```diff
@@ -6,7 +6,16 @@ scalar JSON
 extend type User {
   pledges: [Pledge!]!
   memberships: [Membership!]!
+  "Amount of days left until all active and dormant memberships would expire"
```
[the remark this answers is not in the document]
patrickvenetz: Using `"..."` to have comments in GraphQL-API. While `#` are only visible inside the schema declaration, `"..."` exposes comment to GraphQL-API.

Oddly enough, `#` comment in `servers/republik/graphql/schema-types.js` is printed in GraphQL-API, but not the ones in `servers/republik/modules/crowdfundings/graphql/schema-types.js`. Or in `packages/access/graphql/schema-types.js`, which is first time I used `"..."`.

There's something fishy 🐟 

However, reverting to `#` here.

servers/republik/modules/crowdfundings/graphql/resolvers/_mutations/submitPledge.js
```diff
@@ -47,6 +72,26 @@ module.exports = async (_, args, {pgdb, req, t}) => {
...
+    // Check if options contain valid membership Ids
+    if (
+      pledgeOptions
+        .filter(o => !!(o.customization && o.customization.membershipId))
+        .filter(o => !membershipIds.includes(o.customization.membershipId))
+        .length > 0
+    ) {
+      logger.error(`options must contain valid membership IDs`, { req: req._log(), args, membershipIds })
+      throw new Error(t('api/unexpected'))
+    }
```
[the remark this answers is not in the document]
patrickvenetz: True! It fell like Schuppe from my Augen.

servers/republik/modules/crowdfundings/lib/generateMemberships.js
patrickvenetz: Not inclined to alter `membership.membershipTypeId` as it would alter data into the past with current set of queries.

Suggesting this: `membership.membershipTypeId` becomes a mere reflection of current state of membership and `membershipPeriod.membershipTypeId` sort of keeps track.

Or `membership` describes initial state, latest `membershipPeriod` reflects how it was altered with `membershipTypeId` and `pledgeId` (which caused period). Prefer this one.

servers/republik/modules/crowdfundings/lib/generateMemberships.js
```diff
@@ -60,52 +62,123 @@ module.exports = async (pledgeId, pgdb, t, req, logger = console) => {
...
+          await evaluate({
+            package_: resolvedPackage,
+            packageOption: plo.packageOption,
+            membership
+          })
+
+        await pgdb.public.membershipPeriods.insert(
+          additionalPeriods
+            .map(period => omit(period, ['id', 'createdAt', 'updatedAt']))
+        )
```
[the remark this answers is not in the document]
patrickvenetz: Yes

servers/republik/modules/crowdfundings/lib/generateMemberships.js
patte: I also prefer the later one. I see you already added the `membershipPeriod` <-> `pledgeOption` relationship. I suggest we resolve the (current) type in the [Membership resolver](https://github.com/orbiting/backends/blob/master/servers/republik/modules/crowdfundings/graphql/resolvers/Membership.js#L2) by checking the `membershipTypeId` of the latest period. So it would always be resolved to the current type, but we don't lose information about what it initially was.
We could also add `Membership.initialType` resolver that would do what the current one does. But that's optional now.

servers/republik/modules/crowdfundings/graphql/resolvers/PackageOption.js
```diff
@@ -3,8 +3,7 @@ module.exports = {
     return Promise.all([
```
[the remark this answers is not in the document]
patrickvenetz: `resolvePackages` adopted to fetch goodes and membership types, and skipping querying if `packageOption.reward` is present.

servers/republik/script/prolong/importCancellations.js
```diff
@@ -0,0 +1,97 @@
...
+  if (dry) {
+    console.log("dry run: this won't change anything")
+  }
+
+  const input = rw.readFileSync('/dev/stdin', 'utf8')
+  if (!input || input.length < 4) {
+    throw new Error('You need to provide input on stdin')
+  }
+
+  let stats = {
```
patrickvenetz: (Can be `const`. You're only altering object further below.)

servers/republik/script/prolong/importCancellations.js
```diff
@@ -0,0 +1,97 @@
...
+    numMemberships: 0
+  }
+  const numCancellationsBefore = await pgdb.public.membershipCancellations.count()
+
+  const transaction = await pgdb.transactionBegin()
+  await Promise.all(
```
patrickvenetz: (Maybe use bluebird's `Promise.map`, on which you can steer concurrency. `Promise.all` will execute in parallel.)

servers/republik/script/prolong/importCancellations.js
```diff
@@ -0,0 +1,97 @@
...
+        const normalizedCategory = getNormalizedCategory(category)
+
+        for (let membership of memberships) {
+          await transaction.public.memberships.update(
+            { id: membership.id },
+            { renew: false }
+          )
+          await transaction.public.membershipCancellations.insert({
+            membershipId: membership.id,
+            reason: reason.length ? reason.trim() : null,
```
patrickvenetz: (Maybe attach `category` in it's original to reason, too. That would help understand `OTHER` as in reason there would be "wünscht Papier".)

servers/republik/script/prolong/importCancellations.js
```diff
@@ -0,0 +1,97 @@
...
+  }
+  const numCancellationsBefore = await pgdb.public.membershipCancellations.count()
+
+  const transaction = await pgdb.transactionBegin()
+  await Promise.all(
+    csvParse(input)
+      .map(row => ({
+        ...row,
+        date: moment(row.date, 'DD.MM.YYYY'),
+        userId: row.adminUrl.split('users/')[1]
```
patrickvenetz: (Maybe verify `userId` in database afterwards. Not trusting data in sheet.)

PR #138, opened 2018-11-12
Questionnaires improvements

patte:


packages/voting/graphql/resolvers/QuestionTypeDocumentResult.js
```diff
@@ -1,10 +1,25 @@
...
+    }
     if (!result || !result.path) {
       return null
     }
-    return getDocument(null, { path: result.path }, context)
+    const fields = getFieldList(info, true)
+    const withoutContent = fields.indexOf('content') === -1
+    return search(null, {
+      withoutContent,
+      withoutRelatedDocs: true,
```
tpreusse:
```suggestion
    const relatedDocsNamespaces = [
      'meta.series',
      'meta.format',
      'meta.dossier',
      'meta.discussion'
    ]
    const withoutRelatedDocs = !fields.find(field => (
      relatedDocsNamespaces
        .find(namespace => field.startsWith(namespace))
    ))
    return search(null, {
      withoutContent,
      withoutRelatedDocs,
```

tpreusse: with dataloaders this would not be necessary because we could move the resolve logic to the resolvers! ;)

patte: Agree. And it would potentially speed up querying of multiple documents, as requested docs might point to the same format the dataloader would cache the responce to the same query and only do it once, not for every doc as now. Sadly we didn't and don't have time for cool refactoring projects like this.

PR #142, opened 2018-11-19
WIP cancellation mangement

patte:


servers/republik/migrations/sqls/20181119124113-cancellations-up.sql
```diff
@@ -0,0 +1,30 @@
+CREATE DOMAIN cancel_category AS TEXT
```
patrickvenetz: Uh! This looks more maintainable than `ENUM`. Intended?

servers/republik/modules/crowdfundings/graphql/resolvers/_queries/cancellationCategories.js
```diff
@@ -0,0 +1,17 @@
+const { parse, Source } = require('graphql')
+const Schema = require('../../schema-types')
+
+const cancelCategories = parse(new Source(Schema))
```
patrickvenetz: Maybe `cancellationCategories`

servers/republik/migrations/sqls/20181119124113-cancellations-up.sql
```diff
@@ -0,0 +1,30 @@
...
+  CHECK(
+    VALUE IN ('EDITORIAL', 'NO_TIME', 'VAUNT', 'LOGIN', 'PAPER', 'NO_AUTO_RENEWAL', 'TOO_EXPENSIVE', 'OTHER')
+  );
+
+CREATE TABLE "membershipCancellations" (
+  "id"              uuid primary key not null default uuid_generate_v4(),
+  "membershipId"    uuid not null references "memberships",
```
patrickvenetz:
```
"membershipId"    uuid not null references "memberships"(id),
---------------------------------------------------------^
```

Add `id`, is more explicit.

servers/republik/migrations/sqls/20181119124113-cancellations-up.sql
```diff
@@ -0,0 +1,30 @@
+CREATE DOMAIN cancel_category AS TEXT
+  CHECK(
+    VALUE IN ('EDITORIAL', 'NO_TIME', 'VAUNT', 'LOGIN', 'PAPER', 'NO_AUTO_RENEWAL', 'TOO_EXPENSIVE', 'OTHER')
```
patrickvenetz: `VAUNT` is missing in schema declaration. And quite frankly, not sure what it even is.

servers/republik/migrations/sqls/20181119124113-cancellations-up.sql
```diff
@@ -0,0 +1,30 @@
...
+  );
+
+CREATE TABLE "membershipCancellations" (
+  "id"              uuid primary key not null default uuid_generate_v4(),
+  "membershipId"    uuid not null references "memberships",
+  "reason"          text,
+  "category"        cancel_category not null,
+  "createdAt"       timestamptz default now(),
+  "updatedAt"       timestamptz default now()
```
patrickvenetz: Architectural: Here we might have a chance to keep track when a cancellation was nulled or revoked. Maybe add `revokedAt` and fill when a cancellation becomes `active` again.

It does keep track of cancellations for `MONTHLY_ABO`, too, si?

servers/republik/migrations/sqls/20181119124113-cancellations-up.sql
patte: that's exactly the reason! A domain can easily be changed, enums not.

servers/republik/migrations/sqls/20181119124113-cancellations-up.sql
patte: good catch. `VAUNT` was `Selbstbeweihräucherung`, but it got kicked during the development.

servers/republik/migrations/sqls/20181119124113-cancellations-up.sql
patte: if you say so... ;)

servers/republik/migrations/sqls/20181119124113-cancellations-up.sql
patrickvenetz: If I am not mistaken, there would be only one cancellation record per membership. So, if a membership is reactivate, cancellation record is nilled.

My intend however was to have a cancellation reasons "log".

Say membership A. I cancel in March, Seibt did annoy me. I revoke that cancellation in April, after Philipp persuaded me to do so. Later that year, in November, I want to cancel the membership for good.

`membershipCancellations` table would have two records:

```
membershipId | reason            | created | revoked
-------------|-------------------|---------|--------
membership A | "Seibt ist doof"  | 2018-03 | 2018-04
membership A | "ZEIT ONLINE Abo" | 2018-11 | null
```

Since schedule is tight, we can drop this.

servers/republik/modules/crowdfundings/graphql/schema-types.js
```diff
@@ -260,6 +260,32 @@ type PaymentSource {
   expYear: Int!
 }
 
+enum CancellationCategoryType {
```
patrickvenetz: What went away now is also the "system" cancellation, due to a dormant abo et al. Did like that before. So we'd understand why a `membership.active` flag is set to `false`.

System cancellations could be:
- Viable, dormant membership recognized, cancelling this membership
- Payment did not arrive within expected time period

PR #145, opened 2018-11-22
Delete sessions for stale app instances

garamond:
Fixes https://github.com/orbiting/app/issues/161


packages/notifications/lib/utils.js
```diff
@@ -0,0 +1,11 @@
+const deleteSessionForDevices = async (tokens, pgdb) => {
+  return pgdb.query(`
+    DELETE FROM sessions s WHERE s.id IN (
+                                         SELECT d."sessionId" FROM devices d WHERE d.id IN (:tokens)
```
patte:
```suggestion
      SELECT d."sessionId" FROM devices d WHERE d.id IN (:tokens)
```

packages/notifications/lib/utils.js
```diff
@@ -0,0 +1,11 @@
...
+                                         )
```
patte:
```suggestion
    )
```

packages/notifications/lib/utils.js
patte: I wonder if/why `IN (:tokens)` works. Pogi says you should use `&&` here: https://pogi.readthedocs.io/en/latest/pitfalls/#named-parameter-column-in-expression

garamond: Maybe because of the parens around `:tokens`?

packages/notifications/lib/apn.js
```diff
@@ -58,12 +60,22 @@ const publish = async (args) => {
...
     })
-    debug('sending %d notifications...', tokens.length)
-    return Promise.all(
+    const result = await Promise.all(
       tokens.map(token =>
         provider.send(message, token)
       )
     )
+    debug('APN: #recipients %d, message: %O, result: %O', tokens.length, message, result)
+    const staleTokens = result.failed.reduce((acc, cur) => {
```
tpreusse: where does `result.failed` come from? AFAIK `Promise.all` does not return such a thing.

packages/notifications/lib/utils.js
```diff
@@ -0,0 +1,11 @@
+const deleteSessionForDevices = async (tokens, pgdb) => {
+  return pgdb.query(`
+    DELETE FROM sessions s WHERE s.id IN (
+      SELECT d."sessionId" FROM devices d WHERE ARRAY[d.id] && :tokens
```
tpreusse: should this really be `d.id`? Tokens are saved in the `device.token` column.

packages/notifications/lib/apn.js
garamond: That's true, I have to reduce the `result` array first…

packages/notifications/lib/utils.js
garamond: 👍

PR #149, opened 2018-11-27
Membership extension wave 2

patte:
- [x] user(slug: String, accessToken: ID): User resolver
- [x] update updateMailchimp for wave 2

### DIFF
--- a/packages/auth/graphql/resolvers/_queries/me.js
+++ b/packages/auth/graphql/resolvers/_queries/me.js
@@ -2,11 +2,11 @@ const { getUserByAccessToken } = require('../../../lib/AccessToken')
 
 module.exports = (_, { accessToken }, context) => {
   const { user: me } = context
-  if (me) {
-    return me
-  }
   if (accessToken) {
     return getUserByAccessToken(accessToken, context)
+  }
+  if (me) {
+    return me
   }
   return null
 }
--- a/packages/auth/graphql/resolvers/_queries/user.js
+++ b/packages/auth/graphql/resolvers/_queries/user.js
@@ -1,8 +1,13 @@
 const transformUser = require('../../../lib/transformUser')
 const { resolveUser } = require('../../../lib/Users')
 const Roles = require('../../../lib/Roles')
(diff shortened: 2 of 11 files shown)


tpreusse:
let go with winning `accessToken` on `me` instead support in `user`:

`backends/packages/auth/graphql/resolvers/_queries/me.js`
```diff
const { getUserByAccessToken } = require('../../../lib/AccessToken')

module.exports = (_, { accessToken }, context) => {
  const { user: me } = context
+  if (accessToken) {
+    const user = getUserByAccessToken(accessToken, context)
+    if (user) {
+      return user
+    }
+  }
  if (me) {
    return me
  }
-  if (accessToken) {
-    return getUserByAccessToken(accessToken, context)
-  }
  return null
}
```

I'll review mc script now.

patte:
I generally agree with your feedback (assuming you meant to say `prolongMemberships`. But there are several problems holding me back from a refactor like this:
- You suggest to just cache the memberships of a user passed through `resolveMemberships`. This would be practical in other places too. But then it would be more reusable if `resolveMembership` was a dataloader so it can batch queries and cache the result not bound to a user but generally available.
- me doing the membership evaluation in `updateMailchimp` is not going to be reusable either. We should rather do `User.mustProlongBeforeDate` and `User.canProlongBeforeDate`. Where the first only checks a users active membership and the second checks if there are gifted memberships that can be prolonged. 
- [evaluate](https://github.com/orbiting/backends/blob/master/servers/republik/modules/crowdfundings/lib/CustomPackages/index.js#L74) (which build customPackages) doesn't use [findEligableMemberships]  (https://github.com/orbiting/backends/blob/master/servers/republik/modules/crowdfundings/lib/CustomPackages/index.js#L16) but reproduces the eligibilty checks, this makes it hard to change things without breaking the logic.
Generally I think we should totaly abstain from trying to refactor small parts in the crowdfunding backend and very soon do a broader refactor, where we implement dataloaders for all the involved entities (User, Membership, MembershipType, etc.) and the tree builders `resolveMembership`, `resolvePackage`, `resolveUser`.

PR #151, opened 2018-12-03
Env Powered Discussion Bans

tpreusse:
screen shot 2018-12-03 at 14 35 48

screen shot 2018-12-03 at 14 55 55

For extreme cases where the community team needs to cool someone down.

The banned user will be informed via email. Generally we start with a week or less.

Unrelated change: add empty alt tag to open beacon.


tpreusse:
@patte I understand your concern. You have my commitment that as soon as the 5th person is banned we'll come back to this.

PR #159, opened 2018-12-17
close discussions via document.meta

patte:
- [x] sync doc.meta.discussionClosed to DB
- [x] test it if it works ;)

### DIFF
--- a/packages/discussions/lib/Discussion.js
+++ b/packages/discussions/lib/Discussion.js
@@ -16,6 +16,8 @@ const upsert = async (id, settings = {}, { pgdb, loaders }, legacyDiscussionId) 
     }
   }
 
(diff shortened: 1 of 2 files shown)


packages/discussions/lib/Discussion.js
```diff
@@ -16,6 +16,8 @@ const upsert = async (id, settings = {}, { pgdb, loaders }, legacyDiscussionId)
     }
   }
 
+  const closed = settings.closed || null
```
uxengine: Currently only works one way: You can close a discussion, but not reopen it. After commit and publish, in my article markdown I see `discussionClosed: false` (correct) but the API still delivers `closed: true` for ownDiscussion.

PR #160, opened 2018-12-18
script to remove dialog boxes from published articles

patte:


servers/publikator/script/removeDialogBoxes.js
```diff
@@ -0,0 +1,193 @@
...
+}
+
+const getContext = (payload) => {
+  let loaders = {}
+  const context = {
+    ...payload,
+    loaders,
+    user: {
+      name: 'Publikator-bot',
+      email: 'ruggedly@republik.ch',
```
tpreusse: Use the one that is setup: `ruggedly@project-r.construction`

PR #163, opened 2018-12-20
Paper invoice prolong

patrickvenetz:
run:
`heroku run "cd servers/republik && node script/prolong/paperInvoice.js --dry" -a republik-api --size performance-l > paperInvoice.txt`


servers/republik/modules/crowdfundings/graphql/resolvers/User.js
```diff
@@ -123,6 +123,25 @@ module.exports = {
...
+      const hasPendingPledges = await pgdb.public.query(`
+        SELECT "pledges".* FROM "pledges"
+
+        JOIN "pledgeOptions"
+          ON "pledges"."id" = "pledgeOptions"."pledgeId"
+          AND "pledgeOptions"."membershipId" = :activeMembershipId
+
+        WHERE "pledges"."status" = 'WAITING_FOR_PAYMENT'
+        ;
+      `, { activeMembershipId: memberships.find(m => m.active).id || 0 })
```
patte: Would it not make more sense to move that down to [L161](https://github.com/orbiting/backends/pull/163/files#diff-cee2cfea33e6969e45a1d051cdc6c0e3R161) and use eligableMemberships to find the active one?

servers/republik/script/prolong/paperInvoice.js
```diff
@@ -64,10 +145,14 @@ PgDb.connect().then(async pgdb => {
       ON mt.id = m."membershipTypeId"
     WHERE mt.name = 'BENEFACTOR_ABO' AND
       u.id != :PARKING_USER_ID
```
patte:
```suggestion
       u.id != :PARKING_USER_ID
     ORDER BY
       u."lastName"
```

servers/republik/script/prolong/paperInvoice.js
```diff
@@ -77,12 +162,23 @@ PgDb.connect().then(async pgdb => {
...
```
patte:
```suggestion
       u.id != :PARKING_USER_ID
     ORDER BY
       u."lastName"
```

servers/republik/script/prolong/paperInvoice.js
```diff
@@ -77,12 +162,23 @@ PgDb.connect().then(async pgdb => {
...
 
   log(`investigating ${paperPeople.length} paper people and ${benefactors.length} benefactors`)
 
+  const transaction = await pgdb.transactionBegin()
+
   const benefactorNeedProlong = await enrichWithProlongAndAddress(benefactors)
+    .then(createPledgeAndPayments.bind(
+      this,
+      { transaction, option: PROLONG_PACAKGEOPTION_BENEFACTOR }
+    ))
```
patte: I don't see the point with this bind construct. Why not just map to it?

servers/republik/script/prolong/paperInvoice.js
```diff
@@ -29,33 +30,113 @@ const log = printIds
...
+  id: 'c1cb2b16-8f03-4eec-851c-cea4204d7ff7',
+  price: 24000
+}
+const PROLONG_PACAKGEOPTION_BENEFACTOR = {
+  id: '35be2bea-4d3f-44e8-a224-2d17de6d6d51',
+  price: 100000
+}
+
+const createPledgeAndPayments = async ({ transaction, option }, a) =>
+  Promise.map(a, async d => {
```
patte: what's `a`, `b`?

servers/republik/script/prolong/paperInvoice.js
```diff
@@ -29,33 +30,113 @@ const log = printIds
...
+      membershipId: d.membershipId
+    })
+
+    const payment = await transaction.public.payments.insertAndGet({
+      type: 'PLEDGE',
+      method: 'PAYMENTSLIP',
+      paperInvoice: true,
+      total: option.price,
+      status: 'WAITING',
+      dueDate: moment().add(30, 'days')
```
patte: use a `now` you define at when the script begins to run so that all payments have the same due date. If you are on it, you could also set the `createdAt` of all these entities to now.

servers/republik/modules/crowdfundings/graphql/resolvers/User.js
patrickvenetz: Would keep it there to return early. Thinking: If there is a pending pledge, finding eligible memberships is useless. In fact, we might even move it up a little further - no need to "resolve memberships".

servers/republik/script/prolong/paperInvoice.js
patrickvenetz: Yeah, I'd like to stick to it. Not really a mapping job here, more a contextual one: What transaction, what option to use.

servers/republik/script/prolong/paperInvoice.js
```diff
@@ -120,9 +210,10 @@ PgDb.connect().then(async pgdb => {
     log(csvFormat(paperPeopleNeedProlong.filter(d => d.address).map(mapToCsv)))
     log('---')
   }
+
+  await pgdb.close()
+
   log('finished!')
-}).then(() => {
-  process.exit()
```
patte: at least for me the script doesn't stop now!

servers/republik/script/prolong/paperInvoice.js
```diff
@@ -29,60 +30,162 @@ const log = printIds
...
+        user, {}, { pgdb, user: me }
+      )
+      if (
+        prolongBeforeDate &&
+        moment(prolongBeforeDate).isBefore(PROLONG_BEFORE_DATE)
+      ) {
         const address = await pgdb.public.addresses.findOne({
-          id: user._raw.addressId
+          id: user._raw.addressId,
+          country: ['Schweiz', 'schweiz', 'Liechtenstein']
```
patrickvenetz: @patte Variations in database found thus far:

```
CH
ch
schweiz
SCHWEIZ
Schweiz
Schweiz 
Schweiz / Greece
Suisse
Svizzera
Swiss
Switzerland
Liechtenstein
Liechtenstein 
```

via `SELECT "country" FROM "addresses" GROUP BY "country"`

servers/republik/script/prolong/paperInvoice.js
patrickvenetz: It should not kill the process before `console.log` has printed everything in buffer.
It did before when print CSV lines.

Quitting job manually was easiest.
Graceful quit e.g. with an interval on SIG* would be better.
Use `process.stdout` probably best.

servers/republik/modules/crowdfundings/graphql/resolvers/_mutations/updatePayment.js
```diff
@@ -80,7 +80,13 @@ module.exports = async (_, args, {pgdb, req, t}) => {
...
       }
 
-      if (pledge.total > 100000) {
+      const hasPledgeMemberships = await pgdb.public.memberships.count({
+        pledgeId: pledge.id
+      })
+
+      // Only generate memberships (or periods) of pledge has not generated
+      // memberships already.
+      if (hasPledgeMemberships < 1) {
```
patrickvenetz: This should "fix" `updatePayment` mutation:
- Generate memberships if pledge did not cause any memberships this far e.g. `ABO`
- Generate periods (via `generateMemberships`) if they were not created yet e.g. `PROLONG`

servers/republik/modules/crowdfundings/graphql/resolvers/User.js
patte: but it has to be a prolong pledge to a extendable membership no? anyway, it should always be the same, so it's ok for me.

servers/republik/script/prolong/paperInvoice.js
patte: yes, but for this script, it's enough.

PR #166, opened 2018-12-27
replace phantomjs with chromium/puppeteer on zeit.co's now

patte:
Instead of rendering with phantomJSCloud, (lambdas/chromium/) is deployed as lambda to zeit.co's now. Like before the asset server's `render` endpoint forwards the request to the (lamdba) service and runs the result through the image manipulation pipeline. Due to the lambda needing to check if a url is allowed to be rendered, `RENDER_URL_WHITELIST` was removed from (packages/assets/). Now `URL_WHITELIST` must be provided to chromium-lambda.

Examples:
with cdn: https://cdn.republik.pink/render?width=1200&height=1&url=https%3A%2F%2Frepublik.ch%2F%3FextractId%3DUg9b_wmkL&resize=2000&zoomFactor=2
with asset-server: https://assets.republik.pink/render?width=1200&height=1&url=https%3A%2F%2Frepublik.ch%2F%3FextractId%3DUg9b_wmkL&resize=2000&zoomFactor=2
directly: https://chromium.staging.republik.sh/screenshot?url=https%3A%2F%2Fwww.republik.ch%2F%3FextractId%3DUg9b_wmkL&format=png&zoomFactor=2
share image: https://assets.republik.pink/render?width=1200&height=628&updatedAt=2018-11-22T17%3A59%3A05.884Z&url=https%3A%2F%2Frepublik.love%2Fcommunity%3Fshare%3Db3054752-eefe-4cb4-9da0-b57a9c07d334%26package%3DPROLONG
pdf with charts: https://assets.republik.pink/pdf/2018/09/04/entwerten-die-zentralbanken-unser-geld.pdf

Stress test:
1. goto keycdn.com and purge `rassetsserverstaging`.
2. visit: https://republik.love/2018

## render
the render endpoints newly supports `?viewport=:widthx:height`.
url schema: `/render?url=:url&[viewport=[:width]x[:height]]...`
  - renders :url
  - optional :viewport
    - succeeds deprecated: `width=:width&height=:height`
    - default 1200xAUTO
    - this api always screenshots the full page (with scrolling), cropping to viewport is not supported.
  - optional &fullPage
    - default `1`
    - this api screenshots the full page per default (with scrolling), set `fullPage` to `0` to crop to viewport
  - optional &zoomFactor
    - default 1
  - optional &cookie
    - example: 'id=something'
  - optional &basicAuthUser &basicAuthPass
    - send basic auth when requesting ?url
  - optional &format
    - default `png`
    - supports `png`, `jpeg`
  - optional &quality
    - only for `jpeg`

As the API always renders and screenshots the full page specifying height is not required (can still be handy if the media-queries depend on a reasonable height.)
A higher minimum of height would not result in perfomance gains:
screenshot from 2018-12-28 01-54-30

# Deploy
- deploy assets-api as usual (`git push assets-production`)
- lambda:
```
now secret url_whitelist https://www.republik.ch,https://republik.ch
now -e URL_WHITELIST=@url_whitelist
now alias backends-xxx.now.sh chromium.republik.sh
```
- purge tag `render` for zone `rassetsserver`

# Fixes
- slow phantomJSCloud
- wrong rendered teasers:

old:
phantom2

new:
puppeteer2


packages/assets/README.md
```diff
@@ -20,11 +20,19 @@ Check [assets-backend](https://github.com/orbiting/assets-backend) for a deploya
...
+
+  - renders :url
+  - optional :viewport
+    - succeeds deprecated: `width=:width&height=:height`
+    - default 1200x200
+    - this api always screenshots the full page (with scrolling), cropping to viewport is not supported.
+  - optional :zoomFactor
+    - requires viewport or w/h
+    - default 1
+  - format: png webp conversion not yet supported.
```
tpreusse: outdated

packages/assets/lib/getWidthHeight.js
```diff
@@ -8,8 +8,8 @@ module.exports = (resize) => {
     }
   }
   const [_width, _height] = resize.split('x')
-  const width = _width ? parseInt(_width) : null
-  const height = _height ? parseInt(_height) : null
+  const width = _width ? parseInt(Math.abs(_width)) : null
```
tpreusse: `Math.abs` always return a number or NaN. No need to `parseInt` afterwards.

Generally `parseInt` without radix (2nd param) is [unsafe in browsers](https://coderwall.com/p/4eaixa/parseint-can-be-dangerous). While it's safe in node.js I'd still avoid it, just use the plus operator: `+'0023'`, it behaves as expected in all envs.

packages/assets/lib/renderUrl.js
```diff
@@ -1,59 +1,57 @@
...
+) => {
+  const browser = await puppeteer.connect({
+    browserWSEndpoint: PUPPETEER_WS_ENDPOINT
+  })
+
+  const page = await browser.newPage()
+
+  const promises = [
+    page.setViewport({
+      width: parseInt(Math.abs(width) || 1200),
```
tpreusse: see above, re `Math.abs` returning numbers and `parseInt`

packages/assets/lib/renderUrl.js
```diff
@@ -1,59 +1,57 @@
...
+      height: parseInt(Math.abs(height) || 200),
+      deviceScaleFactor: parseFloat(Math.abs(zoomFactor) || 1)
```
tpreusse: no need for `parseFloat`

packages/assets/README.md
```diff
@@ -20,11 +20,19 @@ Check [assets-backend](https://github.com/orbiting/assets-backend) for a deploya
...
-
-  renders :url with a viewport of :width x :hight to a png. webp conversion not supported.
-
-  ENVs: `PHANTOMJSCLOUD_API_KEY`, `RENDER_URL_WHITELIST`: comma separated, accept: :url.indexOf(whiteUrl) === 0, `FRONTEND_BASIC_AUTH_USER`, `FRONTEND_BASIC_AUTH_PASS`
+- `/render?url=:url&[viewport=[:width]x[:height]]&[zoomFactor=:sf]`
+
+  - renders :url
+  - optional :viewport
+    - succeeds deprecated: `width=:width&height=:height`
+    - default 1200x200
```
tpreusse: I'd prefer `1200x1`. 200 is arbitrary. If the default behaviour is `auto` height, 1 will lead to the right size in all cases except when there are height checks on the rendered pages. However those checks will most-likely also expect more than 200 height.

packages/assets/lib/renderUrl.js
```diff
@@ -1,59 +1,57 @@
...
+  const browser = await puppeteer.connect({
+    browserWSEndpoint: PUPPETEER_WS_ENDPOINT
+  })
+
+  const page = await browser.newPage()
+
+  const promises = [
+    page.setViewport({
+      width: parseInt(Math.abs(width) || 1200),
+      height: parseInt(Math.abs(height) || 200),
```
tpreusse: imo change to 1

packages/assets/lib/getWidthHeight.js
patte: Yes, but the puppeteer api explicitly expects an integer not a number on the width/height input. Without `parseInt` it could also be a float. As float values make no sense here anyway, I thought I add it globally.

packages/assets/README.md
patte: totally agree

packages/assets/lib/getWidthHeight.js
tpreusse: In that case: use `Math.floor` or `Math.ceil`.

lambdas/chromium/screenshot.js
```diff
@@ -0,0 +1,125 @@
...
+      quality,
+      cookie,
+      basicAuthUser,
+      basicAuthPass
+    }
+  } = parse(req.url, true)
+  debug({ url, width, height, zoomFactor, fullPage })
+
+  if (!url) {
+    res.statusCode = 422
```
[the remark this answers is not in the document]
patte: https://stackoverflow.com/questions/3050518/what-http-status-response-code-should-i-use-if-the-request-is-missing-a-required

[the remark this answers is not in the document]
patrickvenetz: Yeah, an opinion rift.

First answer was updated and points to https://www.bennadel.com/blog/2434-http-status-codes-for-invalid-data-400-vs-422.htm

lambdas/chromium/screenshot.js
```diff
@@ -0,0 +1,125 @@
...
+
+    const screenshot = await page.screenshot({
+      fullPage: !(['false', '0'].includes(fullPage)),
+      type,
+      ...quality !== undefined
+        ? { quality: getPosInt(quality) }
+        : {}
+    })
+
+    browser.close()
```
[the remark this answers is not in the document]
patte: Why wait until it's closed?

lambdas/chromium/screenshot.js
```diff
@@ -0,0 +1,125 @@
...
+  }
+}
+
+const getPosInt = (number) =>
+  Math.ceil(Math.abs(number))
+
+// returns buffer
+module.exports = async (req, res) => {
+  const {
+    query: {
```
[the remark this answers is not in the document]
patte: I would rather not do it:
1. It's hard to define boundaris. eg. `height` render can legitimately be > 10000px
1. The error messages generated by chromium (and forwarded to the client) are understandable. Examples
  - options.quality is unsupported for the png screenshots
  - Expected options.quality to be between 0 and 100 (inclusive), got 20000

lambdas/chromium/screenshot.js
```diff
@@ -0,0 +1,131 @@
...
+      basicAuthUser,
+      basicAuthPass
+    }
+  } = parse(req.url, true)
+
+  const format = (_format && FORMATS.indexOf(_format) !== -1)
+    ? _format
+    : 'png'
+
+  debug({ url, width, height, zoomFactor, fullPage, format, quality, cookie, basicAuthUser, basicAuthPass })
```
patrickvenetz:
```suggestion
  debug({ url, width, height, zoomFactor, fullPage, format, quality, cookie, basicAuthUser, basicAuthPass: !!basicAuthPass })
```

Would not print credentials, or at least not password.

servers/assets/README.md
```diff
@@ -1,36 +1,28 @@
 Assets Backend
 ---------------
 
-This is a lightweight, fast nodejs-express server for asset proxying and image manipulation (resizing, greyscaling, webp format transformation). It streams assets from other urls, from s3 buckets, out of github repos and can render webpages to PNGs.
+This is a lightweight, fast nodejs-express server for asset proxying and image manipulation (resizing, greyscaling, webp format transformation). It streams assets from other urls, from s3 buckets, out of github repos and can render via external serves.
```
patrickvenetz:
```suggestion
This is a lightweight, fast nodejs-express server for asset proxying and image manipulation (resizing, greyscaling, webp format transformation). It streams assets from other urls, from s3 buckets, out of github repos and can render via external servers.
```

lambdas/chromium/screenshot.js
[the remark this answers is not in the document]
patrickvenetz: So, not going for 400?

lambdas/chromium/screenshot.js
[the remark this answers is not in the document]
patte: :notes: It's the circle of life... :notes: 
I just didn't want to wait before returning the result, I now changed it accordingly.

lambdas/chromium/screenshot.js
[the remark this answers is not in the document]
patte: You are right, let's go for simple standard.

PR #171, opened 2019-01-07
feat(github): Enable optional GitHub GraphQL API rate limit

patrickvenetz:
This Pull Request limits amount of requests to Github GraphQL.

- No more than 1 concurrent connection
- Wait minimum 200ms before querying again

Uses NPM package [bottleneck](https://www.npmjs.com/package/bottleneck) as a generic job queue. Would allow clustering via Redis.

### DIFF
--- a/packages/github/lib/clients.js
+++ b/packages/github/lib/clients.js
@@ -1,14 +1,32 @@
 const { createApolloFetch } = require('apollo-fetch')
+const Bottleneck = require('bottleneck')
+const debug = require('debug')('github:clients')
+
 const GitHubApi = require('@octokit/rest')
+
 const appAuth = require('./appAuth')
 
 const {
+  GITHUB_GRAPHQL_RATELIMIT,
   GITHUB_LOG_RATELIMIT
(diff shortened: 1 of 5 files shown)


packages/github/lib/clients.js
```diff
@@ -1,15 +1,33 @@
...
 } = process.env
+
 const DEV = process.env.NODE_ENV && process.env.NODE_ENV !== 'production'
 
 let installationToken
 let nextRateLimitCheck
 
+const limiter = new Bottleneck({
+  maxConcurrent: 1,
+  minTime: 200
```
patte:
```suggestion
  minTime: 100
```
Isn't 200 a little bit too slow?

patrickvenetz: Done so. Several test ran flawlessly.

patrickvenetz:
Including Redis datastore now.

Side effect of this Pull Request: Indexing everything works again well. So no more "first `repo` index the `document`index" (which led to GitHub API errors).

PR #172, opened 2019-01-08
Send T-7 email

patrickvenetz:
This Pull Request sends T-7 email to ABO memberships. Other membership types are exempt from this email. Same rule applies to T-30 emails.

It keeps `prolongBeforeDate` query and adds days left as an additional merge variable.

### DIFF
--- a/servers/republik/modules/crowdfundings/lib/Mail.js
+++ b/servers/republik/modules/crowdfundings/lib/Mail.js
@@ -1,4 +1,5 @@
 const debug = require('debug')('crowdfundings:lib:Mail')
+const moment = require('moment')
 
 const { createMail, sendMailTemplate } = require('@orbiting/backend-modules-mail')
 const { grants } = require('@orbiting/backend-modules-access')
@@ -579,6 +580,8 @@ mail.prepareMembershipOwnerNotice = async ({ user, endDate, cancelUntilDate, tem
   const customPledgeToken = AccessToken.generateForUser(user, 'CUSTOM_PLEDGE')
 
   const formattedEndDate = dateFormat(endDate)
+  const timeLeft = moment(endDate).diff(moment())
+  const daysLeft = Math.ceil(moment.duration(timeLeft).as('days'))
 
   return ({
     to: user.email,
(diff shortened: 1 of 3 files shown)


servers/republik/modules/crowdfundings/lib/Mail.js
```diff
@@ -595,6 +598,9 @@ mail.prepareMembershipOwnerNotice = async ({ user, endDate, cancelUntilDate, tem
       { name: 'end_date',
         content: formattedEndDate
       },
+      { name: 'days_left',
+        content: daysLeft >= 1 ? daysLeft : 1
```
patte: I would move that up, with `Math.max(1, Math.ceil(...))`

servers/republik/modules/crowdfundings/lib/scheduler/owners.js
```diff
@@ -58,22 +60,27 @@ const getBuckets = async ({ now }, { pgdb }) => {
...
-      u.id IN (
-        SELECT
-          DISTINCT(m."userId")
-        FROM
-          memberships m
-      )
+      m."userId" != :PARKING_USER_ID
+      AND m.active = true
+      AND m.renew = true
+    GROUP BY 1
```
patte: why did you change `FROM users` to `FROM memberships`? It's not wrong, I just wonder why you made the change to accomplish the same thing, just a little bit slower.

servers/republik/modules/crowdfundings/lib/scheduler/owners.js
```diff
@@ -97,10 +106,24 @@ const getBuckets = async ({ now }, { pgdb }) => {
...
+      stats.numNeedProlongProgress++
 
       if (prolongBeforeDate) {
         const dropped = buckets.some(bucket => {
+          // Don't add user to bucket if user.membershipTypes does not contain
+          // any of memberships listed in bucket.onlyMembershipTypes.
+          if (bucket.onlyMembershipTypes) {
+            const hasNecessaryMembershipType = user.membershipTypes
+              .map(type => bucket.onlyMembershipTypes.includes(type))
+              .reduce((acc, cur) => !cur ? acc : cur, false)
```
patte: just use `!!user.membershipTypes.find...` for more clarity and to stop searching once you found.

servers/republik/modules/crowdfundings/lib/scheduler/owners.js
```diff
@@ -97,10 +106,24 @@ const getBuckets = async ({ now }, { pgdb }) => {
       )
```
patte: `prolongBeforeDate` might be wrong. How can you be sure it's the date for the membershipType you are looking for? I think for this to be correct you need to pass the type to `getProlongBeforeDate` and only check the desired types there. If only we knew these requirements before...

PR #209, opened 2019-02-11
WIP Winback refactor 2

patte:
- edit cancellations


servers/republik/modules/crowdfundings/graphql/resolvers/Membership.js
```diff
@@ -126,11 +127,15 @@ module.exports = {
...
-      })))
+      .then(cancellations => cancellations
+        .map(cancellation => ({
+          ...cancellation,
+          category: {
+            type: cancellation.category
+          }
+        })
+        )
+        .sort((a, b) => ascending(a.createdAt, b.createdAt))
```
patrickvenetz: Would sort in database query wrapper with `orderBy`. It is fair to assume Postgres is faster in sorting.

PR #210, opened 2019-02-12
fix auto login, AUTO_LOGIN -> AUTO_LOGIN_REGEX

patte:
AUTO_LOGIN_REGEX signs in matching emails automatically after 2 seconds.
Set `AUTO_LOGIN_REGEX=^([a-zA-Z0-9._%+-]+)@republik\.ch$` for a more open life in development.

- fix auto `authorizeSession`
- rename AUTO_LOGIN -> AUTO_LOGIN_REGEX

screenshot from 2019-02-12 01-25-44

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,6 @@ env:
     - PUBLIC_WS_URL_PATH=/graphql
     - SEND_MAILS=false
     - FORCE_TEST_END=true
-    - AUTO_LOGIN=true
     - SESSION_SECRET=testtest
     - DEFAULT_MAIL_FROM_NAME='travis'
     - DEFAULT_MAIL_FROM_ADDRESS='travis@test.project-r.construction'
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,6 +44,7 @@ env:
     - PARKING_PLEDGE_ID=00000000-1234-0000-0000-000000000001
     - PARKING_USER_ID=00000000-4321-0000-0000-000000000001
     - DISPLAY_AUTHOR_SECRET=testtest
+    - AUTO_LOGIN_REGEX=^([a-zA-Z0-9._%+-]+)@test\.project-r\.construction$
 before_install:
   - curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version 1.7.0
   - export PATH=$HOME/.yarn/bin:$PATH
--- a/packages/auth/lib/Users.js
+++ b/packages/auth/lib/Users.js
@@ -208,13 +208,12 @@ const signIn = async (_email, context, pgdb, req, consents, _tokenType) => {
 
 const shouldAutoLogin = ({ email }) => {
   if (AUTO_LOGIN_REGEX) {
-    let testMatch = false
     try {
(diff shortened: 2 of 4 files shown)


packages/auth/lib/Users.js
```diff
@@ -205,15 +207,14 @@ const signIn = async (_email, context, pgdb, req, consents, _tokenType) => {
...
-    // - email addresses containing «not» will neither be logged in nor send an sign request
-    const testMatch = email.match(/^([a-zA-Z0-9._%+-]+)@test\.project-r\.construction$/)
-    if (testMatch) {
-      if (testMatch[1].indexOf('not') === -1) {
-        return true
-      }
+  if (AUTO_LOGIN_REGEX) {
+    let testMatch = false
+    try {
+      testMatch = new RegExp(AUTO_LOGIN_REGEX).test(email)
```
patrickvenetz: Maybe `return` early. Don't see (yet) why it should be stuffed into variable.

packages/auth/lib/Users.js
```diff
@@ -205,15 +207,14 @@ const signIn = async (_email, context, pgdb, req, consents, _tokenType) => {
...
+    } catch (e) {
+      testMatch = false
```
patrickvenetz: Maybe `return false` and get rid of `testMatch` variable all together.

packages/auth/lib/Users.js
```diff
@@ -173,11 +173,13 @@ const signIn = async (_email, context, pgdb, req, consents, _tokenType) => {
     })
     if (shouldAutoLogin({ email })) {
       setTimeout(async () => {
-        console.log('AUTO_LOGIN!')
+        console.warn(`🔓💥 auto login for ${email}!!! AUTO_LOGIN_REGEX`)
```
patrickvenetz:
```suggestion
        console.warn(`🔓💥 Auto Login for ${email} due to AUTO_LOGIN_REGEX`)
```

packages/auth/lib/Users.js
```diff
@@ -173,11 +173,13 @@ const signIn = async (_email, context, pgdb, req, consents, _tokenType) => {
...
         await authorizeSession({
           pgdb,
           tokens: [token],
-          email
+          email,
+          req,
```
patrickvenetz: Was this broken before adding `req` and `me`?

.travis.yml
```diff
@@ -44,6 +44,7 @@ env:
     - PARKING_PLEDGE_ID=00000000-1234-0000-0000-000000000001
     - PARKING_USER_ID=00000000-4321-0000-0000-000000000001
     - DISPLAY_AUTHOR_SECRET=testtest
+    - AUTO_LOGIN_REGEX=^([a-zA-Z0-9._%+-]+)@test\.project-r\.construction$
```
patrickvenetz: Remove `AUTO_LOGIN` further above. It appears that `AUTO_LOGIN` is no more.

packages/auth/lib/Users.js
patte: yes, and therefore the tests as well

PR #213, opened 2019-02-19
 feat(crowdfundings): Improve adminUsers search results 

patrickvenetz:
This Pull Request attempts to improve search results on `adminUsers` GraphQL query.

It only changes how database is queried, and what results are returned in which order. It does not improve speed (neither does it decrease it).

**More relevant results**

Query changes relay even more on word similarity and distance calculated by Postgres ("relevance"). The more relevant a term is in on-the-fly generated searchable text (`concat_ws(...)` parts), the further up result should appear.

Results are not shown if their relevance is deemed too low.

There are two thresholds we can adjust: `WORD_SIMILARITY_THRESHOLD` ("similiarity") and `WORD_DISTANCE_THRESHOLD` ("distance"). Results with a similarity ratio below 0.6 are omitted, as well as results whith a distance ratio above 0.5.

Thresholds were set after manual testing with different types of data (membership #, HR-ID) which did deliver relevant results without omitting potential relevant records.

**Prefixes pattern**

Search terms may match multiple data props. Using prefixes helps to precise information one is looking for.

As an example: Searching for membership #1, using "1" as search term would result many results. Searching for "nr:1" however would likely find membership #1 and sort it to top results.

Available prefixes which can precise a search:
* `nr:` to look for a user with a specific membership sequence number (membership #)
* `voucher:` to look for a user with a membership which hold as specific voucher code
* `hrid:` to look for a user with specific payment HR-ID 
* `pspid:` to look for a user with a specific payment service provider ID
* `access:` to look for a user with a specific access grant voucher code 
* `plz:` to look for a user with a specific postal code in addresses

**Other changes**
* Removes query in `paymentSources`. Table is no longer populated.
* Data is paginated in code. This helps to return an precise number of total results without having to run an expensive search query twice. Memory footprint ok.
* Moved filters outside query. Auto-format made a mess.

Fixes https://github.com/orbiting/republik-admin-frontend/issues/33

### DIFF
--- a/servers/republik/modules/crowdfundings/graphql/resolvers/_queries/adminUsers.js
+++ b/servers/republik/modules/crowdfundings/graphql/resolvers/_queries/adminUsers.js
@@ -18,8 +18,20 @@ module.exports = async (
 ) => {
   Roles.ensureUserHasRole(user, 'supporter')
 
-  const filterActive = (dateRangeFilter || stringArrayFilter || booleanFilter)
-  let items = !(search || filterActive)
+  const queryFilters = !!(dateRangeFilter || stringArrayFilter || booleanFilter) &&
+    andFilters([
+      dateRangeFilterWhere(dateRangeFilter, 'u'),
+      stringArrayFilterWhere(stringArrayFilter, 'u'),
+      booleanFilterWhere(booleanFilter, 'u')
+    ])
+
+  const queryOrderBy = search
+    ? 'word_sim, dist DESC'
+    : orderBy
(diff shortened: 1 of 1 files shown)


patte:
One idea came up by reading this: if a search term with a prefix (e.g. `nr:`) is present, it should be easy to only concat / join this information and make the query much faster. But that's out of scope maybe.

patrickvenetz:
Good idea. It may also drop joined tables not used. Expedited into [an enhancement issue](https://github.com/orbiting/backends/issues/214).

PR #226, opened 2019-03-14
feat(mail): Flag to toggle message logging itself

patrickvenetz:
Allows to disable logging a message in `mailLog` table. Disabled logging in mutation `submitInheritenceStory`.

It adds a templates name as a tag for more convenience in Mandrill dashboard.

- [ ] Remove existing message contents in `mailLog` table which was submitted via `submitInheritenceStory` mutation via SQL Stmt.

### DIFF
--- a/packages/crowdsourcing/graphql/resolvers/_mutations/submitInheritanceStory.js
+++ b/packages/crowdsourcing/graphql/resolvers/_mutations/submitInheritanceStory.js
@@ -54,9 +54,15 @@ ${content ? 'Erbgeschichte: ' + content : ''}
 
 `
 
-  return sendMail({
-    to: CS_INHERITANCE_TO,
-    subject: CS_INHERITANCE_SUBJECT,
-    text: emailText
-  }, context)
+  return sendMail(
+    {
+      to: CS_INHERITANCE_TO,
+      subject: CS_INHERITANCE_SUBJECT,
+      text: emailText
+    },
+    context,
+    {
+      logMessage: false
+    }
+  )
 }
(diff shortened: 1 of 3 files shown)


packages/crowdsourcing/graphql/resolvers/_mutations/submitInheritanceStory.js
```diff
@@ -54,9 +54,15 @@ ${content ? 'Erbgeschichte: ' + content : ''}
...
-  }, context)
+  return sendMail(
+    {
+      to: CS_INHERITANCE_TO,
+      subject: CS_INHERITANCE_SUBJECT,
+      text: emailText
+    },
+    context,
+    {
+      logMessage: false
```
patte:
```suggestion
      logMail: false
```
After thinking about it `logMail` feels better than `logMessage`. An email is an email and less a message. Or just shorten it to `log: false`

packages/mail/lib/mailLog.js
```diff
@@ -54,32 +58,40 @@ const send = async ({
...
+    Object.assign(info, log.info)
+  }
+
+  if (template) {
+    info.template = template
+  }
+
+  if (logMessage) {
+    info.message = message
   }
```
patte: Why did you change this code block? What's the advantage of this notation to the old one?

patrickvenetz: Easier to test, read and stitch together.

packages/crowdsourcing/graphql/resolvers/_mutations/submitInheritanceStory.js
patrickvenetz: It still logs whole "email" just not it's content. And since content is wrapped in `message`, would like to stick to it.

PR #232, opened 2019-03-31
feat(access): Add script to generate voucher codes in bulk 

patrickvenetz:
Can generate a series of grants for a user with only voucher codes in campaign. Prefix and amount can be specified, too.

This Pull Request also adds a constraint which requires existing grants to display a campaign. User should only see campaign if grants are in campaign. Grants may be pre-generated e.g. a series
of voucher codes which may be printed on business cards.

Fixes constraint's default `payload`: Allows to create campaigns without any constraints having to generate `payload.slots = {...}`. Previously, schema allowed `Campaign.slots` to be null.

**How to test**

Create a dedicated campaign, using new constraint:

```sql
INSERT INTO "public"."accessCampaigns"("id","title","description","grantPeriodInterval","emailFollowup","beginAt","endAt","constraints","createdAt","updatedAt","grantClaimableInterval")
VALUES
(E'9a7f13da-4e01-4de9-87de-ce4cee00c838',E'Probezugriff über Visitenkarte 🎟',NULL,E'14 days',E'7 days',E'2019-03-30 00:00:00+00',E'2020-01-01 00:00:00+00',E'[{"requireExistingGrants": {}}, {"notGrantable": {}}]',E'2019-03-31 12:54:57.208611+00',E'2019-03-31 12:54:57.208611+00',E'1 year');
```

You should then see in /konto no new campaign. Campaign will show up once your generate grants for logged in user.

Generate voucher code for a user ID, like so:

```sh
packages/access/script/generateVoucherCodes.js --campaign 9a7f13da-4e01-4de9-87de-ce4cee00c838 --granter <Your User ID> --prefix AB --amount 10
```

You should then see in /konto a section showing voucher codes.

### DIFF
--- a/packages/access/lib/constraints/requireExistingGrants.js
+++ b/packages/access/lib/constraints/requireExistingGrants.js
@@ -1,0 +1,51 @@
+const debug = require('debug')('access:lib:constraints:requireExistingGrants')
+
+/**
+ * Contraint checks if a granter has grants in campaign. If false, constraint
+ * will hinder display of campaign.
+ *
+ * Story: User should only see campaign if grants are in campaign. Grants may
+ * be pre-generated e.g. a series of voucher codes which may be printed on
+ * business cards.
+ *
+ * @example: {"requireExistingGrants": {}}
+ */
+
+const getMeta = async (args, context) => {
+  const { granter, campaign } = args
+  const { pgdb } = context
+
+  const hasGrants = await pgdb.query(`
+    SELECT "accessGrants".id
+
+    FROM "accessGrants"
+
+    WHERE
+      "accessGrants"."accessCampaignId" = '${campaign.id}'
+      AND "accessGrants"."granterUserId" = '${granter.id}'
+      AND "accessGrants"."revokedAt" IS NULL
+      AND "accessGrants"."invalidatedAt" IS NULL
+  `)
+
+  const meta = {
+    visible: hasGrants.length > 0,
+    grantable: null,
+    payload: {}
+  }
+
+  debug(
+    'getMeta',
+    {
+      granter: granter.id,
+      campaign,
+      meta
+    }
+  )
+
+  return meta
+}
+
+module.exports = {
+  isGrantable: () => true,
+  getMeta
+}
--- a/packages/access/lib/grants.js
+++ b/packages/access/lib/grants.js
@@ -1,6 +1,7 @@
 const debug = require('debug')('access:lib:grants')
 const moment = require('moment')
 const validator = require('validator')
+const Promise = require('bluebird')
 
 const { Roles } = require('@orbiting/backend-modules-auth')
 const { applyPgInterval: { add: addInterval } } = require('@orbiting/backend-modules-utils')
@@ -10,6 +11,8 @@ const constraints = require('./constraints')
 const eventsLib = require('./events')
 const mailLib = require('./mail')
 const membershipsLib = require('./memberships')
+
+const VOUCHER_CODE_LENGTH = 5
 
 const evaluateConstraints = async (granter, campaign, email, t, pgdb) => {
   const errors = []
@@ -45,6 +48,24 @@ const evaluateConstraints = async (granter, campaign, email, t, pgdb) => {
   debug({ errors })
 
   return { errors }
+}
+
+const generate = async (granter, campaignId, grants = [], pgdb) => {
+  const campaign = await campaignsLib.findOne(campaignId, pgdb)
+
+  if (campaign === undefined) {
+    throw new Error('Campaign not found.')
+  }
+
+  await Promise.map(grants, async fields => {
+    fields.granterUserId = granter.id
+    fields.accessCampaignId = campaign.id
+    fields.beginBefore = addInterval(moment(), campaign.grantClaimableInterval)
+
+    const grant = await pgdb.public.accessGrants.insertAndGet(fields)
+
+    await eventsLib.log(grant, 'generate', pgdb)
+  }, { concurrency: 10 })
 }
 
 const grant = async (granter, campaignId, email, message, t, pgdb, mail) => {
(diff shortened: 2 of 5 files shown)


packages/access/lib/grants.js
```diff
@@ -47,6 +50,24 @@ const evaluateConstraints = async (granter, campaign, email, t, pgdb) => {
   return { errors }
 }
 
+const generate = async (granter, campaignId, grants = [], pgdb) => {
```
patte: This function doesn't generate anything, IMHO `insert` would be a more appropriate name.

packages/access/lib/constraints/requireExistingGrants.js
```diff
@@ -0,0 +1,51 @@
...
+
+    WHERE
+      "accessGrants"."accessCampaignId" = '${campaign.id}'
+      AND "accessGrants"."granterUserId" = '${granter.id}'
+      AND "accessGrants"."revokedAt" IS NULL
+      AND "accessGrants"."invalidatedAt" IS NULL
+  `)
+
+  const meta = {
+    visible: hasGrants.length > 0,
```
patte: what does this mean for existing/future global campaings? Would they no tbe invisble to everyone initially?

patrickvenetz: Constraints are bound per campaign via `campaigns.constraints`.

Some examples:

Campaign A's constraints:
```
[
  {
    "requireRole": {
      "roles": [
        "editor",
        "supporter",
        "admin"
      ]
    }
  },
  {
    "limitSlots": {
      "slots": 100
    }
  },
  {
    "limitRevokedSlots": {
      "slots": 100
    }
  },
  {
    "recipientInNoSlot": {}
  },
  {
    "recipientNotGranter": {}
  }
]
```

Campaign B's constraints:
```
[
  {
    "requireExistingGrants": {}
  },
  {
    "notGrantable": {}
  }
]
```

Further, by design `false` will trump `true` grantable and visible flags. So if there is one constraint returning `false` it won't be grantable or visible (aka. returned).

packages/access/lib/grants.js
patrickvenetz: Yes, `insert` is better.

packages/access/lib/grants.js
```diff
@@ -47,6 +50,24 @@ const evaluateConstraints = async (granter, campaign, email, t, pgdb) => {
...
+  const campaign = await campaignsLib.findOne(campaignId, pgdb)
+
+  if (campaign === undefined) {
+    throw new Error('Campaign not found.')
+  }
+
+  await Promise.map(grants, async fields => {
+    fields.granterUserId = granter.id
+    fields.accessCampaignId = campaign.id
+    fields.beginBefore = addInterval(moment(), campaign.grantClaimableInterval)
```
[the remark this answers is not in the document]
patrickvenetz: Very true! Thanks. Refactored.

packages/access/script/generateVoucherCodes.js
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    const grants = []
+
+    let iteration = 0
+    while (grants.length < argv.amount && iteration < 5 * argv.amount) {
+      iteration++
+
+      const grant = {
+        voucherCode: await pgdb.queryOneField(`SELECT make_hrid('"accessGrants"'::regclass, 'voucherCode'::text, ${grantsLib.VOUCHER_CODE_LENGTH}::bigint)`)
```
patte: why not use the transaction here? Then you could skip the check if it's already in the table on L71.

packages/access/script/generateVoucherCodes.js
```diff
@@ -0,0 +1,94 @@
...
+      }
+
+      if (argv.prefix) {
+        grant.voucherCode = argv.prefix + grant.voucherCode.slice(argv.prefix.length, grantsLib.VOUCHER_CODE_LENGTH)
+      }
+
+      if (
+        await pgdb.public.accessGrants.count({ voucherCode: grant.voucherCode }) === 0 &&
```
patte:
```suggestion
```

packages/access/script/generateVoucherCodes.js
```diff
@@ -0,0 +1,94 @@
...
+  })
+  .help()
+  .version()
+  .argv
+
+PgDb.connect().then(async pgdb => {
+  const transaction = await pgdb.transactionBegin()
+
+  try {
+    const granter = await pgdb.public.users.findOne({ id: argv.granter })
```
patte:
```suggestion
    const granter = await transaction.public.users.findOne({ id: argv.granter })
```

packages/access/lib/grants.js
```diff
@@ -47,6 +50,25 @@ const evaluateConstraints = async (granter, campaign, email, t, pgdb) => {
...
+    const grant = await pgdb.public.accessGrants.insertAndGet({
+      ...fields,
+      granterUserId: granter.id,
+      accessCampaignId: campaign.id,
+      beginBefore: addInterval(moment(), campaign.grantClaimableInterval)
+    })
+
+    await eventsLib.log(grant, 'insert', pgdb)
+  }, { concurrency: 10 })
+}
```
patte: looking at it now, I don't completely understand why you dont map the grants into a new array and then insert this with one `pgdb.insert`?

packages/access/script/generateVoucherCodes.js
patrickvenetz: `grant.voucherCode` can be altered after `make_hrid`, namely added prefix on L67.

But could look-up voucher code only of `argv.prefix` is set.

patte: :man_facepalming: sorry I missed this part when re-reading it.

packages/access/script/generateVoucherCodes.js
```diff
@@ -0,0 +1,110 @@
...
+
+    debug({ grants })
+
+    if (grants.length < argv.amount) {
+      throw new Error(`Unable to generate ${argv.amount} unqiue voucher codes. Change parameters.`)
+    }
+
+    console.log(`inserting ${grants.length} grant(s)...`)
+
+    // await grantsLib.insert(granter, argv.campaign, grants, transaction)
```
patte: :warning:

PR #258, opened 2019-05-14
refactor(base): graphiql instead of graphql-playground

patte:
graphiql is prefered because of its url query support and cleaner ui
files are included locally to be able to develop offline


packages/base/express/graphiql.js
```diff
@@ -0,0 +1,10 @@
+const express = require('express')
+const path = require('path')
+
+module.exports = (server) => {
+  server.use(
+    '/graphiql',
+    express.static(path.join(__dirname, './graphiql/'))
```
tpreusse:
```js
express.static(path.join(__dirname, 'graphiql'), {
  redirect: false
})
```

Currently we enforce `/graphiql/?query=` urls, which are ugly. Alt would be `{index: false}` and making index.html only available under `/graphiql` (without slash)

P.S.: Dark bright mix is ugly

patte: Agreed! Without the `/` is better, I will do it.
I also agree that the bright dark mix is ugly. I wanted the dark mode because graphiql was the only dev-tool that I use regularly that wasn't dark and thus the contrast hurt my eyes. And because I didn't want to spend too much time on this I went with just an (codemirror) editor dark theme. I think it's already better than with the default theme. But you frontend-wizards are very welcome to smoothen out the experience :wink:

PR #260, opened 2019-05-16
feat(assets): chache render results in s3

patte:


packages/assets/express/render.js
```diff
@@ -1,16 +1,34 @@
...
+
+const { AWS_S3_BUCKET } = process.env
+const cacheS3Path = `render-cache/`
+
+const getCacheKey = (params) => {
+  const paramsString = Object.keys(params)
+    .sort(ascending)
+    .map(key => `${key}:${params[key]}`)
+    .join('::')
+  return crypto
```
tpreusse: Some human-readable info in the filename would be potentially helpful. E.g.
```suggestion
  const sha = crypto.createHash('sha256')
    .update(paramsString)
    .digest('hex')
  const s3KeyEncode = string => encodeURIComponent(string).replace(/%/g, 'C')
  const url = (params.url || '').replace('https://', '')
  return `${s3KeyEncode(url).slice(0, 500)}-${sha}`
```

Max length of an [s3 key is 1024 chars](https://stackoverflow.com/questions/6870824/what-is-the-maximum-length-of-a-filename-in-s3) including path.

PR #261, opened 2019-05-16
feat(documents): optimise linked/related docs loading

tpreusse:
only load linked docs in returned content or children

This speeds up getting the front document a lot. Locally the response time goes from ~140ms to ~70ms (both with hot caches) with the smaller dev documents collection. In prd the gains may reach 5x to 10x. I expect the current prd ~700ms to go down to ~200ms.

~ToDo~:
- ~new `afterDate` & `beforeDate` query for `Document.children` for overview page~

Decided not to implement advanced children filtering in backend (credit line based) and instead hard code some teaser ids in the frontend: https://github.com/orbiting/republik-frontend/commit/7bdab34cc56717c146003fdc5f75e16c3285dd58#diff-048df1fa60145e7b472e7ed8cafbcf72R37

This will allow for the same optimisations.

### DIFF
--- a/packages/assets/lib/webp.js
+++ b/packages/assets/lib/webp.js
@@ -3,6 +3,11 @@ const addSuffix = (url) => {
   if (!query) {
     return base
   }
+  // prevent double suffix
+  // - e.g. when resolving twice (content & children)
+  if (base.endsWith('.webp')) {
+    return url
+  }
   return `${base}.webp?${query}`
 }
 
--- a/packages/search/graphql/resolvers/_queries/search.js
+++ b/packages/search/graphql/resolvers/_queries/search.js
@@ -161,7 +161,7 @@ const createHighlight = (indicesList) => {
 
 const defaultExcludes = [ 'contentString', 'resolved' ]
 const createQuery = (
-  searchTerm, filter, sort, indicesList, user, scheduledAt, withoutContent, withoutAggs, ignorePrepublished
+  searchTerm, filter, sort, indicesList, user, scheduledAt, withoutChildren, withoutAggs, ignorePrepublished
 ) => ({
   query: {
     bool: {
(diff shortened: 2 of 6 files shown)


packages/search/lib/Documents.js
```diff
@@ -200,55 +200,103 @@ const getElasticDoc = (
...
+          contextRepoId
+        )
+      }
+    }
+    const repoId = getRepoId(node.url, 'autoSlug')
+    if (repoId) {
+      repos.push(repoId)
+    }
+  })
+  return { repos, users }
```
patte:
```suggestion
  return {
    repos: repos.filter(Boolean),
    users: users.filter(Boolean)
  }
```

packages/assets/lib/webp.js
```diff
@@ -3,6 +3,11 @@ const addSuffix = (url) => {
   if (!query) {
     return base
   }
+  // prevent double suffix
+  // - e.g. when resolving twice (content & children)
+  if (base.endsWith('.webp')) {
+    return url
+  }
```
patte: :+1:

packages/search/graphql/resolvers/_queries/search.js
```diff
@@ -354,12 +357,14 @@ const search = async (__, args, context, info) => {
...
-  let withoutContent
-  if (info) {
-    withoutContent = _withoutContent !== undefined ||
-      !hasFieldRequested('Document.content', info)
-  } else {
-    withoutContent = _withoutContent || false
+  let withoutContent = _withoutContent || false
+  let withoutChildren = withoutContent
+  if (info && _withoutContent === undefined) {
+    withoutContent = !hasFieldRequested('content', info)
```
patte: Wouldn't this __wired__ query trigger content even though it's not needed?
`{ documents {nodes {meta {ownDiscussion {comments {nodes {content}}}}}}}`

tpreusse: Probably. Maybe `hasFieldRequested` could be extended to support a max depth but would require a unit test approach to ensure correct behaviour between `search`, `documents` and `document`. IMO ignore for now since unlikely. What do you think?

PR #264, opened 2019-05-22
feat(employees): shuffle and boost

tpreusse:
New argument for `employees` to shuffle and limit the result to the specified count
 - one famous female and one famous male is boosted
 - ensures unique users

### DIFF
--- a/servers/republik/graphql/resolvers/_queries/employees.js
+++ b/servers/republik/graphql/resolvers/_queries/employees.js
@@ -1,14 +1,32 @@
+const { shuffle } = require('d3-array')
 const { transformUser } = require('@orbiting/backend-modules-auth')
 
-module.exports = async (_, args, {pgdb}) => {
-  const data = await pgdb.public.gsheets.findOneFieldOnly({name: 'employees'}, 'data')
+module.exports = async (_, args, { pgdb }) => {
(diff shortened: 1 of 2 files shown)


servers/republik/graphql/resolvers/_queries/employees.js
```diff
@@ -1,14 +1,32 @@
...
+
+    result = [
+      boosted[0],
+      rest[0],
+      boosted[1]
+    ].concat(rest.slice(1))
+      .filter(Boolean)
+      .filter((a, i, all) => all.findIndex(b => b.userId === a.userId) === i)
+      .slice(0, args.shuffle)
+  }
```
patte: Wouldn't it be nicer if the boosted users appear in random positions? Thus shuffling the result array again in the end?

tpreusse: Current use case is to request 6 and display 3 on mobile and 6 on desktop. Generally I think it makes sense to put boosted ones in the beginning. In any display those are very visible spots.

patte: Alright if the sample is that small it makes perfect sense.

tpreusse:
@patte thanks for the review

PR #277, opened 2019-06-25
feat(auth): Add an email code authentication challenge

patrickvenetz:
ToDo:
- [x] "Rate-limit" `authorizeSession` attempts
- [x] A sensible set of tests

### DIFF
--- a/packages/auth/__tests__/auth.int.jest.js
+++ b/packages/auth/__tests__/auth.int.jest.js
@@ -293,7 +293,7 @@ describe('sign in with tokenType EMAIL_CODE', () => {
   })
 })
 
-describe('authorizeSession', () => {
+describe.only('authorizeSession', () => {
   test('rate limit attempts within session', async () => {
     const resultSignIn = await signIn({
       user: Users.Unverified,
@@ -313,13 +313,20 @@ describe('authorizeSession', () => {
       expect(resultAuthorize.data).toBeFalsy()
     }
 
-    const resultAuthorize = await authorizeSession({
+    const rateLimitedAuthorize = await authorizeSession({
       email: Users.Unverified.email,
       tokens: [{ type: 'EMAIL_TOKEN', payload: resultSignIn.payload }],
       apolloFetch
     })
 
-    expect(resultAuthorize.data).toBeFalsy()
+    expect(rateLimitedAuthorize.data).toBeFalsy()
+    expect(rateLimitedAuthorize.errors).toEqual(
+      expect.arrayContaining([
+        expect.objectContaining({
+          message: expect.stringMatching(/^Zu viele Anmeldeversuche/)
+        })
+      ])
+    )
   })
 
   test('rate limit attempts in multiple sessions', async () => {
@@ -347,6 +354,13 @@ describe('authorizeSession', () => {
     })
 
     expect(rateLimitedAuthorize.data).toBeFalsy()
+    expect(rateLimitedAuthorize.errors).toEqual(
+      expect.arrayContaining([
+        expect.objectContaining({
+          message: expect.stringMatching(/^Zu viele Anmeldeversuche/)
+        })
+      ])
+    )
   })
 })
 
--- a/packages/auth/__tests__/auth.int.jest.js
+++ b/packages/auth/__tests__/auth.int.jest.js
@@ -1,5 +1,4 @@
 const { Instance } = require('@orbiting/backend-modules-test')
-
 const OTP = require('otp')
 const {
   signIn,
@@ -206,7 +205,7 @@ test('sign in with tokenType APP default tokenType', async () => {
   expect(resultSession2.signInResult.data.signIn.tokenType).toBe('EMAIL_TOKEN')
 })
 
-describe(`sign in with tokenType EMAIL_CODE`, () => {
+describe('sign in with tokenType EMAIL_CODE', () => {
   const tokenType = 'EMAIL_CODE'
   const MAX_VALID_TOKENS = 5
 
@@ -225,7 +224,6 @@ describe(`sign in with tokenType EMAIL_CODE`, () => {
   })
 
   test('denies authorization with different session', async () => {
-
     const resultSignIn = await signIn({ user: Users.Unverified, tokenType, skipAuthorization: true })
     expect(resultSignIn.signInResult.data.signIn.phrase).toBeTruthy()
     expect(resultSignIn.signInResult.errors).toBeFalsy()
@@ -240,7 +238,6 @@ describe(`sign in with tokenType EMAIL_CODE`, () => {
   })
 
   test('fails if amount of simultaniously valid tokens is exceeded', async () => {
-
     for (let i = 0; i < MAX_VALID_TOKENS; i++) {
       await signIn({
         user: Users.Unverified,
@@ -253,7 +250,7 @@ describe(`sign in with tokenType EMAIL_CODE`, () => {
     expect(resultSignIn.signInResult.errors).toBeTruthy()
   })
 
-  test('fails if amount of simultaniously valid tokens in multiple is exceeded', async () => {
+  test('fails if amount of simultaniously valid tokens in multiple sessions is exceeded', async () => {
     for (let i = 0; i < MAX_VALID_TOKENS; i++) {
       await signIn({
         user: Users.Unverified,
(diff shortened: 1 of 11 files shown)


packages/auth/__tests__/auth.int.jest.js
```diff
@@ -206,6 +206,75 @@ test('sign in with tokenType APP default tokenType', async () => {
...
+      email: Users.Unverified.email,
+      tokens: [{ type: tokenType, payload: resultSignIn.payload }],
+      apolloFetch: resultSignIn.apolloFetch
+    })
+
+    expect(resultAuthorize.data && resultAuthorize.data.authorizeSession).toBeTruthy()
+  })
+
+  test('denies authorization with different session', async () => {
+    const tokenType = 'EMAIL_CODE'
```
patte:
```suggestion
```

packages/auth/__tests__/auth.int.jest.js
```diff
@@ -206,6 +206,75 @@ test('sign in with tokenType APP default tokenType', async () => {
...
+      email: Users.Unverified.email,
+      tokens: [{ type: tokenType, payload: resultSignIn.payload }],
+      apolloFetch: global.instance.createApolloFetch()
+    })
+
+    expect(resultAuthorize.data).toBeFalsy()
+  })
+
+  test('fails if amount of simultaniously valid tokens is exceeded', async () => {
+    const tokenType = 'EMAIL_CODE'
```
patte:
```suggestion
```

packages/auth/__tests__/auth.int.jest.js
```diff
@@ -206,6 +206,75 @@ test('sign in with tokenType APP default tokenType', async () => {
...
+    for (let i = 0; i < MAX_VALID_TOKENS; i++) {
+      await signIn({
+        user: Users.Unverified,
+        tokenType,
+        skipAuthorization: true
+      })
+    }
+
+    const resultSignIn = await signIn({ user: Users.Unverified, tokenType, skipAuthorization: true })
+    expect(resultSignIn.signInResult.errors).toBeTruthy()
```
patte: how is it possible to recover from this situtation? and could it be tested also?

packages/auth/lib/Sessions.js
```diff
@@ -69,7 +69,7 @@ const sessionByToken = async ({ pgdb, token, email: emailFromQuery }) => {
     `, token)
 
   if (sessions && sessions.length > 0) {
-    if (sessions[0].sess.email !== emailFromQuery) {
+    if (sessions[0].sess.email.toLowerCase() !== emailFromQuery.toLowerCase()) {
```
patte:
```suggestion
    if (!emailFromQuery || sessions[0].sess.email.toLowerCase() !== emailFromQuery.toLowerCase()) {
```

patte: why is this neccesarry when you don't lowercase in [auth/lib/Users.js](https://github.com/orbiting/backends/pull/277/files#diff-0eeae7f2c73b01d9b94ce4e17b94977cR296)?

packages/auth/lib/Users.js
```diff
@@ -276,6 +283,38 @@ const denySession = async ({ pgdb, token, email: emailFromQuery, me }) => {
...
+    )
+
+    if (session.sess.authorizeAttempts > MAX_AUTHORIZE_ATTEMPTS) {
+      throw new AuthorizationRateLimitError({
+        email: emailFromQuery,
+        authorizeAttempts: session.sess.authorizeAttempts
+      })
+    }
+  })
+
```
patte: querying sessions and updating them should go into a transaction. As it's very possible that there is another simultanious authorizeSession running that is doing the same, or one that is currently authorizing the session. Like this they would overwrite each other without anybody noticing.
Not being able to update a session without querying it first (due deep JSON update) is the root of this problem, which I also don't see a general solution to right now.
I would go for a transaction and `select for update`.

packages/auth/lib/challenges/EmailCodeChallenge.js
```diff
@@ -0,0 +1,82 @@
...
+      console.error('Unable to generate a new token: Found too many valid tokens.')
+      throw new TokensExceededError({ email, tokens: existingPayloads, MAX_VALID_TOKENS })
+    }
+
+    let attempts = 10 ** CODE_LENGTH
+    let payload = null
+
+    do {
+      payload = `${Math.round(Math.random() * (10 ** CODE_LENGTH))}`
+        .slice(0, CODE_LENGTH)
```
patte:
```suggestion
```

packages/auth/lib/challenges/EmailCodeChallenge.js
```diff
@@ -0,0 +1,82 @@
...
+        type: Type,
+        email,
+        'expiresAt >=': new Date()
+      }
+    ))
+      .map(token => token.payload)
+
+    if (existingPayloads.length >= MAX_VALID_TOKENS) {
+      console.error('Unable to generate a new token: Found too many valid tokens.')
+      throw new TokensExceededError({ email, tokens: existingPayloads, MAX_VALID_TOKENS })
```
patte:
```suggestion
      throw new TokensExceededError({ email, MAX_VALID_TOKENS })
```

packages/auth/__tests__/auth.int.jest.js
patrickvenetz: It is possible to recover by waiting for one or more tokens to expire.

This could be tested, too. :+1:

packages/auth/lib/Sessions.js
patrickvenetz: An email in `sessions[0].sess.email` can be a user entered input and is case sensitive. Same for  `emailFromQuery`.

Meanwhile [auth/lib/Users.js](https://github.com/orbiting/backends/pull/277/files#diff-0eeae7f2c73b01d9b94ce4e17b94977cR296) is case insensitive.

packages/auth/lib/challenges/EmailCodeChallenge.js
patrickvenetz: Would like to keep this as `Math.random()`. It may return `1` and produce `CODE_LENGTH` + 1 digits. Padding won't clip numbers.

```js
> `${Math.random() * (10 ** 6)}`.slice(0, 6).padStart(6, 0)
'892769'
> `${1 * (10 ** 6)}`.slice(0, 6).padStart(6, 0)
'100000'
> `${Math.random() * (10 ** 6)}`.padStart(6, 0)
'673722'
> `${1 * (10 ** 6)}`.padStart(6, 0)
'1000000'
       ^------- too long
```

packages/auth/__tests__/auth.int.jest.js
patrickvenetz: Tests added. (Also some more.)

packages/auth/__tests__/auth.int.jest.js
```diff
@@ -206,6 +205,151 @@ test('sign in with tokenType APP default tokenType', async () => {
...
+      expect(resultAuthorize.data).toBeFalsy()
+    }
+
+    const resultAuthorize = await authorizeSession({
+      email: Users.Unverified.email,
+      tokens: [{ type: 'EMAIL_TOKEN', payload: resultSignIn.payload }],
+      apolloFetch
+    })
+
+    expect(resultAuthorize.data).toBeFalsy()
```
patte: It's not clear to me what the end state of this test is. the last attempt is also denied, or does it go through? Also how does one recover from too many failed attempts to authrorize a session?

packages/auth/lib/challenges/EmailCodeChallenge.js
```diff
@@ -0,0 +1,82 @@
...
+
+    do {
+      payload = `${Math.round(Math.random() * (10 ** CODE_LENGTH))}`
+        .slice(0, CODE_LENGTH)
+        .padStart(CODE_LENGTH, 0)
+
+      attempts--
+    } while (existingPayloads.includes(payload) && attempts > 0)
+
+    if (attempts === 0) {
```
patte:
```suggestion
    if (existingPayloads.includes(payload) && attempts === 0) {
```
otherwise the last try to generate a payload has succeeded

packages/auth/__tests__/auth.int.jest.js
patrickvenetz: Last attempt won't go through, it does "rate limit attempts within session" (named test that way). Error returned is a generic `api/token/invalid` error message to not give away anything, but could assert that so it is more explicit.

A user can not recover from a rate limited session by an action, but will have to wait until either of the tokens linked to rate limited sessions is expired. [Expired tokens won't be taken into account when auditing authorize attempts.](https://github.com/orbiting/backends/blob/2d0a086c2b396afef9b2217c40dc9ed8394eb2b7/packages/auth/lib/Users.js#L287-L298)

packages/auth/lib/Users.js
```diff
@@ -275,7 +280,53 @@ const denySession = async ({ pgdb, token, email: emailFromQuery, me }) => {
...
+          authorizeAttempts: session.sess.authorizeAttempts,
+          maxAttempts
+        })
+      }
+    })
+
+    await transaction.transactionCommit()
+  } catch (e) {
+    await transaction.transactionRollback()
+    throw e
```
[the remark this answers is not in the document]
patrickvenetz: Fair enough. Will return an audit authorize error and log details into console.

packages/auth/__tests__/auth.int.jest.js
patte: I see. The session is not found in this query if **all** the tokens have expired, thus it's not evaluated for `authorizeAttempts`. That makes sense. As this ratelimiter is not scoped to EMAIL_CODE I would go for an explicit error like "Anzahl Anmelde versuche überschritten, bitte warten Sie xx minuten." So that the careteam and we can see if people bump into this.

patrickvenetz: Adopting.

PR #300, opened 2019-08-15
Revision of 44 transactional emails

eightam:


patrickvenetz:
(Closed in favour of an upcoming, conflict resolved branch since code base changed since August 15.)

PR #304, opened 2019-08-27
refactor(notifications): more permissive upsertDevice, deprecate rollDeviceToken, cleanup deprecated definition

patte:
- [x] avoid floding the log with `duplicate key value violates unique constraint "devices_sessionId_key"` at `upsertDevice` mutation
- [x] deprecated rollDeviceToken
- [x] cleanup deprecated definition in graphql schemas


packages/notifications/graphql/schema.js
```diff
@@ -5,9 +5,11 @@ schema {
 
 type mutations {
   upsertDevice(token: ID!, information: DeviceInformationInput!): Device!
-  rollDeviceToken(oldToken: String!, newToken: String!): Device!
 
-  # user's can remove their devices
+  # deprecated: not used in app anymore, evict if no API calls tracked anymore
```
patrickvenetz: There is this beautiful directive in GraphQL schemas: `@deprecated`. I can see this used here instead of having deprecation notice in a comment.

```
rollDeviceToken(oldToken: String, newToken: String!): Device! @deprecated(reason: "Lorem ipsum dolor sit amet")
```

Bildschirmfoto 2019-08-28 um 05 55 17

As far as I can tell it bears no impact on querying or mutating but is soley for [documentaton and tooling purposes](https://graphql.github.io/graphql-spec/June2018/#sec--deprecated). It even allows Markdown formatted string in `reason` argument 🌟

patte: Thanks for pointing it out. I tried to adopt this style when the definition of deprecated fields changed in graphql sometime last year. Sadly this directive is not available for enums nor mutation params, which let me to use the description. For the sake of beauty I now adopted it where possible.

PR #315, opened 2019-09-12
feat(publikator): script to split a front repo (date based)

patte:


tpreusse:
It worked and is finally live. Thank you @patte

PR #331, opened 2019-11-08
feat(access): Accept an optional payload while requesting or claiming…

patrickvenetz:
… an access grant

* Schema changes validate allowed data in payload.
* Payload can receive a document ID, repo ID, a varation and position string.

### DIFF
--- a/packages/access/graphql/schema.js
+++ b/packages/access/graphql/schema.js
@@ -27,7 +27,7 @@ type mutations {
   """
   claimAccess(
     voucherCode: String!
-    payload: AccessGrantPayloadInput
+    payload: JSON
   ): AccessGrant!
 
   """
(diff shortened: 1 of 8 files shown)


packages/access/graphql/schema.js
```diff
@@ -25,14 +25,18 @@ type mutations {
   """
   Claim a granted membership with a voucher code
   """
-  claimAccess(voucherCode: String!): AccessGrant!
+  claimAccess(
+    voucherCode: String!
+    payload: AccessGrantPayloadInput
```
tpreusse:
```suggestion
    payload: JSON
```
I would just accept arbitrary json. This would allow us to start tracking new stuff without a backend change. E.g. add referrers or other stuff as we see fit. And since the whole object is optional there is no guarantee anyhow.

packages/access/graphql/schema.js
```diff
@@ -25,14 +25,18 @@ type mutations {
...
+  ): AccessGrant!
 
   """
   Request access for one-self
   """
   requestAccess(
     "An ID of an existing AccessCampaign"
     campaignId: ID!
+    payload: AccessGrantPayloadInput
```
tpreusse: same here

PR #332, opened 2019-11-13
fix(auth): prevent payload collision on authorize and deny

tpreusse:
Happened with type EMAIL_CODE. These payload are only unique per email address. Other types use uuids which are unlikely to collide.


patrickvenetz:
LGTM

PR #333, opened 2019-11-19
Notice email about automated charge to membership owners

patrickvenetz:
This Pull Request sends earliest 5 days before a membership is planned to be renewed a transactional email, notify a user about upcoming charge.

Code is run [in membership owner scheduler](https://github.com/orbiting/backends/compare/feat-autopay-notice-5?expand=1#diff-7fd8e0d8a72be494c664c0b3c22c555d) in which we hold now a new bucket ("membership_owner_autopay_notice_5"), and altered scheduler to allow `autoPay` to be a bucket condition.

Membership owner scheduler [runs every 10 minutes](https://github.com/orbiting/backends/compare/feat-autopay-notice-5?expand=1#diff-588ec577936693da893ff2fc8d4b4424R43-R51) instead of once a day.

Upcoming charge is determined using [last membership pledge and it's (most recent) payment](https://github.com/orbiting/backends/compare/feat-autopay-notice-5?expand=1#diff-21d5bc1d878369209a6114907a73e222). This is subject to change depending on upcoming business logic.

Credit card details are [retrieved from payment service provider payload in a payment](https://github.com/orbiting/backends/compare/feat-autopay-notice-5?expand=1#diff-21d5bc1d878369209a6114907a73e222R39-R47) (`payment.pspPayload`) and passed along.

### DIFF
--- a/packages/utils/findLastMembershipPledge.js
+++ b/packages/utils/findLastMembershipPledge.js
@@ -35,14 +35,13 @@ module.exports = async (membershipId, pgdb) => {
     { orderBy: { createdAt: 'DESC' }, limit: 1 }
   )
 
-  // Used card in payment (if available)
-  const card =
-    payment.pspPayload &&
-    payment.pspPayload.source &&
-    payment.pspPayload.source.card
-
   return {
-    ...pledge,
-    card
+    pledge,
+    autoPayPreferences: {
+      card: payment.pspPayload &&
+        payment.pspPayload.source &&
+        payment.pspPayload.source.card,
+      total: pledge.total
+    }
   }
 }
--- a/packages/utils/findLastMembershipPledge.js
+++ b/packages/utils/findLastMembershipPledge.js
@@ -6,7 +6,7 @@ module.exports = async (membershipId, pgdb) => {
     return false
   }
 
-  // Find pledgeOptions, in which autoPay-flag is set
+  // Find pledgeOptions
   const relatedPledgeOptions = await pgdb.public.pledgeOptions.find(
     {
       or: [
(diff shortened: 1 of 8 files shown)


packages/utils/findLastMembershipPledge.js
```diff
@@ -0,0 +1,48 @@
...
+    }
+  )
+
+  if (relatedPledgeOptions.length < 1) {
+    return false
+  }
+
+  // Find latest pledge
+  const pledge = await pgdb.public.pledges.findOne(
+    { id: relatedPledgeOptions.map(po => po.pledgeId), status: 'SUCCESSFUL' },
```
tpreusse: limit to same user id as membership

packages/utils/findLastMembershipPledge.js
```diff
@@ -0,0 +1,48 @@
...
+  )
+
+  // Used card in payment (if available)
+  const card =
+    payment.pspPayload &&
+    payment.pspPayload.source &&
+    payment.pspPayload.source.card
+
+  return {
+    ...pledge,
```
tpreusse: add new `autoPayValue`:
- load all pledge options from the latest pledge
- find pledge option for this membership
- if this is the only pledge option use pledge total otherwise use option total

packages/utils/findLastMembershipPledge.js
tpreusse: check needs to happen earlier that either T-30 or T-5 auto pay is sent

packages/utils/findLastMembershipPledge.js
patrickvenetz: Noted, adopting this in a subsequent Pull Request.

packages/utils/findLastMembershipPledge.js
```diff
@@ -0,0 +1,47 @@
+module.exports = async (membershipId, pgdb) => {
+  // Find membership
+  const membership = await pgdb.public.memberships.findOne({ id: membershipId })
+
+  if (!membership) {
+    return false
+  }
+
+  // Find pledgeOptions, in which autoPay-flag is set
```
tpreusse: rm outdated comment

patrickvenetz:
giphy (1)

PR #341, opened 2019-11-30
Enable TABLEBOOK pledge

patrickvenetz:
This Pull Request enables us to handle Goodie-only custom packages.

It also adds a custom package rule, `endTooFarAway`, which won't pass if there is no active membership or an active membership ends within a "horizon" date.

This Pull Request includes two new email templates dedicated to a pledge named `TABLEBOOK`.

Also, `CUSTOM_PLEDGE` access token scope expands to pledge package `TABLEBOOK`, too, and results in no regressions (e.g. already issues tokens still work).

**SQL**

```sql
BEGIN;
INSERT INTO "public"."packages"("id","crowdfundingId","name","createdAt","updatedAt","companyId","paymentMethods","isAutoActivateUserMembership","custom","rules","order","group")
VALUES
(E'ca911d63-04d7-4b87-94a0-b21cb55a75d5',E'fb546aaa-09e0-403e-8f42-c6ee2f9b437e',E'TABLEBOOK',E'2019-11-26 12:06:16.627189+00',E'2019-11-26 12:06:16.627189+00',E'240ef27d-cf26-48c1-81df-54b2a10732f4',E'{STRIPE,POSTFINANCECARD,PAYPAL,PAYMENTSLIP}',FALSE,TRUE,E'["endTooFarAway"]',110,E'ME');
INSERT INTO "public"."packageOptions"("id","packageId","rewardId","minAmount","maxAmount","defaultAmount","price","userPrice","minUserPrice","createdAt","updatedAt","vat","order","disabled")
VALUES
(E'e39dcd24-802e-479d-89ac-daf5f9d0d45f',E'ca911d63-04d7-4b87-94a0-b21cb55a75d5',E'e2aeb033-38fd-46b0-aadd-f4f4211bd75a',1,10,1,5800,FALSE,0,E'2019-11-26 12:07:08.219594+00',E'2019-11-26 12:07:08.219594+00',0,150,FALSE);
COMMIT;
```

### DIFF
--- a/servers/republik/modules/crowdfundings/lib/CustomPackages/rules/earlyAdopterBonus.js
+++ b/servers/republik/modules/crowdfundings/lib/CustomPackages/rules/earlyAdopterBonus.js
@@ -13,12 +13,12 @@ module.exports = ({ package_, packageOption, membership, payload, now }) => {
       'membership.createdAt not within early adopter range (< %s)',
       MEMBERSHIP_CREATED_BEFORE
     )
-    return
+    return true
   }
 
   if (membership.pledge.userId !== package_.user.id) {
     debug('membership not pledged by user. rule does not apply.')
-    return
+    return true
   }
 
   if (!PACKAGE_ELIGABLE.includes(membership.pledge.package.name)) {
@@ -26,7 +26,7 @@ module.exports = ({ package_, packageOption, membership, payload, now }) => {
       'package "%s" is not eligable for early adopter bonus',
       membership.pledge.package.name
     )
-    return
+    return true
   }
 
   if (payload.additionalPeriods.length === 0) {
@@ -34,7 +34,7 @@ module.exports = ({ package_, packageOption, membership, payload, now }) => {
       'no additional periods found. rule does not apply.',
       payload.additionalPeriods.length
     )
-    return
+    return true
   }
 
   if (membership.membershipPeriods.length >= 3) {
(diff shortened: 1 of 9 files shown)


servers/republik/modules/crowdfundings/lib/CustomPackages/rules/endTooFarAway.js
```diff
@@ -0,0 +1,33 @@
+const debug = require('debug')('crowdfundings:lib:CustomPackages:rule:endTooFarAway')
+
+const { getLastEndDate } = require('../../utils')
+
+const HORIZON_MONTHS = 3
```
tpreusse: Let's ensure all march people have to prolong too.
```suggestion
const HORIZON_MONTHS = 4
```

servers/republik/modules/crowdfundings/lib/CustomPackages/rules/earlyAdopterBonus.js
```diff
@@ -13,36 +13,36 @@ module.exports = ({ package_, packageOption, membership, payload, now }) => {
       'membership.createdAt not within early adopter range (< %s)',
       MEMBERSHIP_CREATED_BEFORE
     )
-    return
+    return true
```
tpreusse: this rule always return true because it mutates and never restricts? Maybe add a comment at the top that this is a mutating rule that always passes

patrickvenetz: Added [some more comments](https://github.com/orbiting/backends/pull/341/files#diff-925a20658e774e9b2028dac18c6ae8c2R219-R235) on rule handling ("passing", "mutating") in code parts which evaluate rules.

PR #342, opened 2019-11-30
Sections leftovers

patrickvenetz:
This Pull Request wires sections feature into all parts of republik, publikator servers.

### DIFF
--- a/packages/documents/lib/resolve.js
+++ b/packages/documents/lib/resolve.js
@@ -221,6 +221,7 @@ const metaFieldResolver = (meta, allDocuments = [], errors) => {
     series,
     dossier: resolver(meta.dossier),
     format: resolver(meta.format),
+    section: resolver(meta.section),
     discussion: resolver(meta.discussion)
   }
 }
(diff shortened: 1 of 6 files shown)


packages/search/lib/Documents.js
```diff
@@ -330,12 +330,17 @@ const addRelatedDocs = async ({
       const extractedIds = extractIdsFromNode(doc.content, doc.repoId)
       userIds = userIds.concat(extractedIds.users)
       repoIds = repoIds.concat(extractedIds.repos)
+    } else {
+      const extractedIds = extractIdsFromNode({ children: doc.meta.credits }, doc.repoId)
```
tpreusse: thanks, I messed this up when optimising and never got around to fixing.

packages/documents/lib/resolve.js
```diff
@@ -76,7 +76,7 @@ const extractUserUrl = url => {
     return
   }
   return extractUserPath(
-    `${urlObject.path}${urlObject.hash || ''}`
+    `${urlObject.pathname}${urlObject.hash || ''}`
```
tpreusse: even though it currently does not matter, let's keep the convention that everything after `hostname` is available for resolving / extracting:
```suggestion
    `${urlObject.pathname}${urlObject.search}${urlObject.hash}`
```

PR #349, opened 2019-12-18
fix Membership- and RevenueStats caching

patte:
- resuse previous caching lib
- scheduler to keep cache warm


servers/republik/modules/crowdfundings/lib/scheduler/index.js
```diff
@@ -75,6 +78,22 @@ const init = async (_context) => {
...
+      runFunc: (args, context) =>
+        Promise.all([
+          surplus(null, { min: '2019-12-01', forceRecache: true }, context),
+          evolution(null, { min: '2019-12', max: '2020-03', forceRecache: true }, context)
+        ]),
+      lockTtlSecs: 6,
+      runIntervalSecs: 8
+    })
+  )
+  //
```
patrickvenetz:
```suggestion
```

PR #358, opened 2020-01-06
MailLog API and Mandrill Webhook

patrickvenetz:
This Pull Request exposes mail log data and accepts Mandrill webhook calls.

Data can be queried:
- `mailLog` for latest transactional emails sent
- `User.mailLog` for latest transactional emails sent, linked to a user.

To query data in a more performant way, some indices were added.

It also accepts signed calls on `/mail/mandrill/webhook` and it's payload is stored - if it can be found - records in `mailLog.mandrillLastEvent`.

Relates to https://github.com/orbiting/republik-admin-frontend/pull/51


packages/auth/loaders/User.js
```diff
@@ -8,6 +10,29 @@ module.exports = (context) => ({
...
+
+      if (or.length === 0) {
+        return []
+      }
+
+      return context.pgdb.public.users.find({ or }).then(users => users.map(transformUser))
+    },
+    null,
+    (array, rows) => rows.find(row => {
+      const values = array.filter(Boolean).map(value => value.toLowerCase())
```
patte: this line is constant for `array, rows`. move out of `rows.find`

packages/mail/graphql/resolvers/MailLogRecord.js
```diff
@@ -0,0 +1,37 @@
...
+  status: record =>
+    (record.mandrillLastEvent && record.mandrillLastEvent.msg.state) ||
+    (record.result && record.result.status) ||
+    record.status.toLowerCase(),
+  error: record =>
+    (record.mandrillLastEvent && record.mandrillLastEvent.msg && record.mandrillLastEvent.msg.diag) ||
+    (record.result && record.result.reject_reason) ||
+    (record.error && record.error.message) ||
+    (record.error && record.error.meta && record.error.meta.error && record.error.meta.error.type),
+  user: async (record, args, { loaders }) => loaders.User.byIdOrEmail.load([record.userId, record.email]),
```
patte:
```suggestion
  user: async (record, args, { loaders }) => loaders.User.byIdOrEmail.load(record.userId || record.email),
```
would this not be enough and more deterministic. If there is a `record.userId` the email was sent to the user no matter what email the user now has, if there is no `userId` get the current user with the `email`

packages/mail/graphql/resolvers/MailLogRecord.js
```diff
@@ -0,0 +1,37 @@
...
+  mandrill: (record, args, { auth, user: me }) => {
+    if (!auth.Roles.userIsInRoles(me, ['admin', 'supporter'])) {
+      return null
+    }
+
+    return [
+      record.createdAt > moment().subtract(30, 'days') && record.result && record.result._id && {
```
patte:
```suggestion
      moment(record.createdAt).isAfter(moment().subtract(30, 'days')) && record.result && record.result._id && {
```

packages/mail/graphql/resolvers/User.js
```diff
@@ -0,0 +1,18 @@
...
+    }
+
+    const records = await pgdb.public.mailLog.find({
+      or: [
+        { id: user.id },
+        { email: user.email }
+      ]
+    }, { orderBy: { createdAt: 'DESC' } })
+
+    return paginator({ first: 40, ...args }, a => a, () => records)
```
patte: as this is available to everybody, limit first to 100.

servers/republik/server.js
```diff
@@ -92,15 +96,15 @@ const run = async (workerId, config) => {
 
   if (LOCAL_ASSETS_SERVER) {
     const { express } = require('@orbiting/backend-modules-assets')
-    for (let key of Object.keys(express)) {
+    for (const key of Object.keys(express)) {
```
patte: :ok_hand:

servers/republik/server.js
```diff
@@ -111,7 +115,8 @@ const run = async (workerId, config) => {
       ...defaultContext,
       t,
       signInHooks,
-      mail,
+      mail: crowdfundingsMail,
+      auth,
```
patte: I'm not a fan of this, as it adds an inconstitency in the usage of the critical auth code. Searching the code for import of `require('@orbiting/backend-modules-auth')` doesn't work anymore. If you really want to get down that route you should at least include it in the `server/publikator/server.js` as well, so that it's available in all contextes. Actually you should include it in all context creations (tests).

PR #359, opened 2020-01-06
feat(mail): Send some transactionals via nodemailer

patrickvenetz:
This Pull Request enables to use to send emails via [nodemailer](https://nodemailer.com/about/) instead of Mandrill: `NodemailerInterface`

`NodemailerInterface` also replaces simple handlebars variables.

To test, setup these ENV variables first, here including some examples.
```sh
# If Regular Expression is matched on a template name, nodemailer is used to sent email
SEND_MAILS_NODEMAILER_REGEX=^signin.*

# Connection URL to e.g. SMTP account nodemailer uses.
SEND_MAILS_NODEMAILER_CONNECTION_URL='smtps://inbox@domain.tld:lengthy-password-with-dashes@asmtp.mail.hostpoint.ch/?pool=true'
```

Upcoming changes include:
- Allow to pause sending over either interface e. g. to pause Mandrill
- Script to resume paused emails

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -104,6 +104,13 @@ SEND_MAILS=false
 # see https://mandrill.zendesk.com/hc/en-us/articles/205582467-How-to-Use-Tags-in-Mandrill
 SEND_MAILS_TAGS=dev,localhost
 
+# If Regular Expression is matched on a template name, nodemailer is used to sent email
+SEND_MAILS_NODEMAILER_REGEX=
+
+# Connection URL to e.g. SMTP account nodemailer uses.
+SEND_MAILS_NODEMAILER_CONNECTION_URL=
+
+
 # required for mails to work, set SEND_MAILS to false for a quick start
 #MANDRILL_API_KEY=
 
--- a/.env.example
+++ b/.env.example
@@ -105,11 +105,10 @@ SEND_MAILS=false
 SEND_MAILS_TAGS=dev,localhost
 
 # If Regular Expression is matched on a template name, nodemailer is used to sent email
-SEND_MAILS_NODEMAILER_REGEX=
+#SEND_MAILS_NODEMAILER_REGEX=^signin*
 
 # Connection URL to e.g. SMTP account nodemailer uses.
-SEND_MAILS_NODEMAILER_CONNECTION_URL=
-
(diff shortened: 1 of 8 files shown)


packages/mail/NodemailerInterface.js
```diff
@@ -0,0 +1,59 @@
+const nodemailer = require('nodemailer')
+
+const { SendMailError } = require('./errors')
+
+const transporter = nodemailer.createTransport(process.env.SEND_MAILS_NODEMAILER_CONNECTION_URL)
```
patte: suggestion: only define `transporter` if `SEND_MAILS_NODEMAILER_CONNECTION_URL` is defined and check in `isUsable` if `transporter` is defined

.env.example
```diff
@@ -104,6 +104,13 @@ SEND_MAILS=false
 # see https://mandrill.zendesk.com/hc/en-us/articles/205582467-How-to-Use-Tags-in-Mandrill
 SEND_MAILS_TAGS=dev,localhost
 
+# If Regular Expression is matched on a template name, nodemailer is used to sent email
+SEND_MAILS_NODEMAILER_REGEX=
+
+# Connection URL to e.g. SMTP account nodemailer uses.
+SEND_MAILS_NODEMAILER_CONNECTION_URL=
```
patte: an example URL would be nice

packages/mail/MandrillInterface.js
```diff
@@ -26,6 +26,9 @@ const MandrillInterface = ({ logger }) => {
         })
       })
     },
+    isUsable () {
+      return true
```
patte:
```suggestion
      return MANDRILL_API_KEY && MANDRILL_API_KEY.length
```
then [checkEnv](https://github.com/orbiting/backends/pull/359/files#diff-00ad47e61ae2ecc80a14cd4c2ff9fbfbR10) could be removed

packages/mail/NodemailerInterface.js
```diff
@@ -0,0 +1,59 @@
...
+              (template, mergeVar) => {
+                const { name, content } = mergeVar
+                return template.replace(new RegExp(`{?{{ ?${name} ?}}}?`, 'ig'), content)
+              },
+              message.text
+            ),
+            html: message.html && variables.reduce(
+              (template, mergeVar) => {
+                const { name, content } = mergeVar
+                return template.replace(new RegExp(`{?{{ ?${name} ?}}}?`, 'ig'), content)
```
patte: DRY variable replacement

PR #362, opened 2020-01-08
feat(mail): Schedule some emails instead of sending them

patrickvenetz:
This Pull Request can deem emails to be "scheduled" and be placed in `mailLog` as such.

ENV variable `SEND_MAILS_SCHEDULE_REGEX` accepts a regular expression and matches mail template names against it; if true, an email is scheduled instead of send.

Due to factors unbeknownst to us, we may need to hold off on sending emails but still have them queued. `mailLog` with records on status `SCHEDULED` just to that.

To test, setup ENV variable first, here including some examples.

```sh
# If Regular Expression is matched on a template name, mail is schedule instead of sent
SEND_MAILS_SCHEDULE_REGEX=^[^signin].*
```

Upcoming changes include:
- Script to sent previously scheduled emails

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -110,6 +110,9 @@ SEND_MAILS_TAGS=dev,localhost
 # Connection URL to e.g. SMTP account nodemailer uses.
 #SEND_MAILS_NODEMAILER_CONNECTION_URL='smtps://inbox@domain.tld:lengthy-password-with-dashes@asmtp.mail.hostpoint.ch/?pool=true'
 
+# If Regular Expression is matched on a template name, mail is schedule instead of sent
+#SEND_MAILS_SCHEDULE_REGEX=^[^signin].*
+
 # required for mails to work, set SEND_MAILS to false for a quick start
 #MANDRILL_API_KEY=
 
--- a/.env.example
+++ b/.env.example
@@ -105,13 +105,13 @@ SEND_MAILS=false
 SEND_MAILS_TAGS=dev,localhost
 
 # If Regular Expression is matched on a template name, nodemailer is used to sent email
-#SEND_MAILS_NODEMAILER_REGEX=^signin*
+#SEND_MAILS_NODEMAILER_TEMPLATE_REGEX=^signin*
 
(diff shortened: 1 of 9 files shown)


packages/mail/utils/shouldScheduleMessage.js
```diff
@@ -0,0 +1,30 @@
...
+
+const DEV = NODE_ENV && NODE_ENV !== 'production'
+
+module.exports = (mail, message) => {
+  const { SEND_MAILS_LOG = true } = process.env
+  const logger = SEND_MAILS_LOG && SEND_MAILS_LOG !== 'false'
+    ? console.log
+    : a => {}
+
+  // don't send in dev, expect SEND_MAILS is true
```
patte:
```suggestion
  // don't schedule in dev, expect SEND_MAILS is true
```

packages/mail/utils/shouldScheduleMessage.js
```diff
@@ -0,0 +1,30 @@
...
+  // don't send mails if SEND_MAILS is false
```
patte:
```suggestion
  // don't schedule mails if SEND_MAILS is false
```

.env.example
```diff
@@ -110,6 +110,9 @@ SEND_MAILS_TAGS=dev,localhost
 # Connection URL to e.g. SMTP account nodemailer uses.
 #SEND_MAILS_NODEMAILER_CONNECTION_URL='smtps://inbox@domain.tld:lengthy-password-with-dashes@asmtp.mail.hostpoint.ch/?pool=true'
 
+# If Regular Expression is matched on a template name, mail is schedule instead of sent
+#SEND_MAILS_SCHEDULE_REGEX=^[^signin].*
```
patte:
```suggestion
#SEND_MAILS_SCHEDULE_TEMPLATE_REGEX=^[^signin].*
```
to distinguish it from email-address regex. applies to `SEND_MAILS_NODEMAILER_REGEX` too, sorry I missed it there.

patrickvenetz: Yeah, true. This env variable names keep growing and growing and growing.

PR #366, opened 2020-01-15
Autopay: Fallout

patrickvenetz:
- Won't send emails if charge attempt was anything but `type: 'card_error'`. These errors will have to be checked manually.
- Auto-Pay related mail templates updated with a (temporary) apology
- Restore membership-owners scheduler, and limit amount of users looked at per run (`MEMBERSHIP_SCHEDULER_USER_LIMIT`)

### DIFF
--- a/servers/republik/modules/crowdfundings/lib/Mail.js
+++ b/servers/republik/modules/crowdfundings/lib/Mail.js
@@ -457,6 +457,15 @@ mail.prepareMembershipOwnerNotice = async ({ user, endDate, graceEndDate, cancel
 }
 
 mail.sendMembershipOwnerAutoPay = async ({ autoPay, payload, pgdb, t }) => {
+  // Ditch sending email if it's not a card_error
+  if (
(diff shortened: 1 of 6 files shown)


servers/republik/modules/crowdfundings/lib/Mail.js
```diff
@@ -457,6 +457,15 @@ mail.prepareMembershipOwnerNotice = async ({ user, endDate, graceEndDate, cancel
 }
 
 mail.sendMembershipOwnerAutoPay = async ({ autoPay, payload, pgdb, t }) => {
+  // Ditch sending email if it's not a card_error
+  if (
+    payload.chargeAttemptStatus !== 'SUCCESS' &&
+    payload.chargeAttemptError &&
+    (!payload.chargeAttemptError.raw || payload.chargeAttemptError.raw.type !== 'card_error')
+  ) {
+    return
```
patte: maybe log to slack?

PR #368, opened 2020-01-16
move embeds into their own package, fix intervalScheduler

patte:
extracts embeds code from publikator into it's own package. Side effect: the embeds query is also available on the republik API, but as it's protected by a roles check it's ok IMHO.

fixes interval scheduler not releasing lock


packages/embeds/lib/index.js
```diff
@@ -0,0 +1,17 @@
...
+const twitter = require('./twitter')
+const vimeo = require('./vimeo')
+const youtube = require('./youtube')
+
+module.exports = {
+  documentcloud: require('./documentcloud'),
+  twitter,
+  vimeo,
+  youtube,
+  imageKeys: uniq([
```
patrickvenetz: Fun fact, I remember a vanilla way to uniqify an array with `Set`:

```js
const bueaechstabjini = [...new Set(['a', 'b', 'c', 'c'])]
```

packages/embeds/package.json
```diff
@@ -0,0 +1,26 @@
...
+  "author": "Patrick Recher <patrick.recher@republik.ch>",
+  "license": "AGPL-3.0",
+  "publishConfig": {
+    "access": "public"
+  },
+  "bugs": {
+    "url": "https://github.com/orbiting/backends/issues"
+  },
+  "homepage": "https://github.com/orbiting/backends#readme",
+  "dependencies": {
```
patrickvenetz: I find it comforting and useful when adding all external package dependencies (`... require(...)`) to `package.json`. (`@orbiting/...` not necessary, there is always one version which yarn handles.) Will help when exporting either module as a standalone version. Until then it comes in super-handy when upgrading packages (e.g. `yarn upgrade-interactive --latest`) and works swiftly as a documentation.

patte: I just forgot, thanks for spotting it.

PR #374, opened 2020-02-12
feat(crowdfundings): Delete unused membership when cancelling a pledge

patrickvenetz:
This Pull Request [will delete a membership entity](https://github.com/orbiting/backends/compare/feat-delete-unused-memberships?expand=1#diff-69972cdee095623d2759b03411dcf759R122) if there is no period to be found in mutation `cancelPledge`.

`evaluatedPeriods` returns an array of periods (and what should happen to each period upon pledge cancellation). It returns an empty array if there is no period to be found.


patrickvenetz:
Some unit tests in `servers/republik/lib/fakeUUID.u.jest.js` are failing since error message changed. But I am unable to reproduce that locally. Any hints?

```
FAIL servers/republik/lib/fakeUUID.u.jest.js
  ● fakeUUID() › faulty argument throws TypeError
    expect(received).toThrowErrorMatchingSnapshot()
    Snapshot name: `fakeUUID() faulty argument throws TypeError 1`
    Snapshot: "The \"data\" argument must be one of type string, Buffer, TypedArray, or DataView. Received type number"
    Received: "The \"data\" argument must be of type string or an instance of Buffer, TypedArray, or DataView. Received type number (123)"
      23 | 
      24 |   test('faulty argument throws TypeError', () => {
    > 25 |     expect(() => UUT(123)).toThrowErrorMatchingSnapshot()
         |                            ^
      26 |     expect(() => UUT({ foo: 'bar' })).toThrowErrorMatchingSnapshot()
      27 |   })
      28 | })
      at Object.<anonymous> (servers/republik/lib/fakeUUID.u.jest.js:25:28)
```

patrickvenetz:
Error messages changed in recent minor node versions which lead to snapshot error in unit tests above.

PR #386, opened 2020-03-09
add missing indices

patte:
add missing indices based on most time consuming, most frequently invoked and slowest execution time of postgres diagnose.


patrickvenetz:
Is there a chance to verify before/after?

patte:
@patrickvenetz 
Screenshot from 2020-03-10 16-27-33
Screenshot from 2020-03-10 16-27-00
Screenshot from 2020-03-10 16-26-41
Screenshot from 2020-03-10 16-26-21

patte:
slow me queries are now 2.5x faster
Screenshot from 2020-03-10 16-35-50
Screenshot from 2020-03-10 16-33-36

PR #394, opened 2020-03-16
removed all kampa references, back to normal (except for pot-membersh…

clarajeanne:
…ip stuff)


patrickvenetz:
thx @clarajeanne

PR #403, opened 2020-04-05
feat(documents): Merge repo URL search params before resolving document URL

patrickvenetz:
This Pull Requests changes return type of `getRepoId(url, requireQuery)`, and merges search parameters of repo URL with resolved document URL.

**Return type of `getRepoId(url, requireQuery)`**

Instead of a string, it returns an object with these props:

- `repoId`: set if repo ID is parseable
- `parsedUrl`: returns an `URL` object with parsed argument `url`

Changing return type here reduces `new URL` calls on same URL.

**Merge search parameters of repo URL with resolved document URL**

Keeps repo URL search parameters when resolving to a document URL.

_Example_

https://github.com/republik-dev/pae-article-zielartikel?autoSlug&foo=bar&audio=1#irgend-ein-hash

becomes

https://frontend.url/2022/02/22/web-slug?foo=bar&audio=1#irgend-ein-hash

### DIFF
--- a/packages/documents/lib/resolve.js
+++ b/packages/documents/lib/resolve.js
@@ -22,29 +22,34 @@ const getRepoId = (url, requireQuery) => {
   ])
 
   if (!url) {
-    return
+    return {}
   }
-  const {
-    hostname,
-    pathname,
-    searchParams
-  } = new URL(String(url), FRONTEND_BASE_URL)
+
+  const parsedUrl = new URL(String(url), FRONTEND_BASE_URL)
+  const { hostname, pathname, searchParams } = parsedUrl
+
   if (!pathname) { // empty for mailto
-    return
+    return {}
   }
+
   const pathSegments = pathname.split('/').filter(Boolean)
   if (
     hostname !== 'github.com' ||
     pathSegments.length !== 2 ||
     !GITHUB_ORGS.split(',').includes(pathSegments[0])
   ) {
-    return
+    return {}
   }
+
   if (requireQuery && !searchParams.has(requireQuery)) {
-    return
+    return {}
   }
+
+  searchParams.delete(requireQuery)
+
   pathSegments[0] = GITHUB_LOGIN
-  return pathSegments.join('/')
+
+  return { repoId: pathSegments.join('/'), parsedUrl }
 }
 
 const userPath = /^\/~([^/?#]+)/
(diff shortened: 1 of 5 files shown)


packages/documents/lib/resolve.js
```diff
@@ -22,29 +22,34 @@ const getRepoId = (url, requireQuery) => {
...
-    return
+    return { parsedUrl }
   }
+
   if (requireQuery && !searchParams.has(requireQuery)) {
-    return
+    return { parsedUrl }
   }
+
+  searchParams.delete(requireQuery)
```
patte, line 48: ~`searchParams` is not used again, why bother modifying it here?~ I see you use it on the returned `parsedUrl`. Could you add a comment explaining this. Also why does require query need to deleted on it anyway?

packages/documents/lib/resolve.js
```diff
@@ -97,28 +102,52 @@ const createUrlReplacer = (allDocuments = [], usernames = [], errors = [], urlPr
...
   if (linkedDoc) {
-    const hash = url.split('#')[1]
-    return `${urlPrefix}${linkedDoc.content.meta.path}${searchString}${hash ? `#${hash}` : ''}`
+    // Stitch and parse simple URL version including arguments {urlPrefix}, {searchString}
+    const resolvedUrl = new URL(
+      `${urlPrefix}${linkedDoc.content.meta.path}${searchString}`,
+      FRONTEND_BASE_URL
+    )
+
+    // Add {parsedUrl.hash} to {resolvedUrl.hash}, incase there is one provided.
```
patte, line 126:
```suggestion
    // Replace {parsedUrl.hash} with {resolvedUrl.hash}
```

packages/documents/lib/resolve.js
```diff
@@ -97,28 +102,52 @@ const createUrlReplacer = (allDocuments = [], usernames = [], errors = [], urlPr
...
   const linkedDoc = allDocuments
     .find(d => d.meta.repoId === repoId)
+
   if (linkedDoc) {
-    const hash = url.split('#')[1]
-    return `${urlPrefix}${linkedDoc.content.meta.path}${searchString}${hash ? `#${hash}` : ''}`
+    // Stitch and parse simple URL version including arguments {urlPrefix}, {searchString}
+    const resolvedUrl = new URL(
+      `${urlPrefix}${linkedDoc.content.meta.path}${searchString}`,
+      FRONTEND_BASE_URL
```
patte, line 123: `FRONTEND_BASE_URL` wasn't involved here before. Are you sure it doesn't have any unwanted consequences?

patrickvenetz: Yes, as sure as automated and manual testing made me.

packages/documents/lib/resolve.js
patrickvenetz: `requireQuery` is either not set or set to `autoSlug` (thus far).

`autoSlug` in URL search params is more of a "flow" property ("do resolve this URL") and less one that should be propagated.

Using `searchParams.delete` is misleading. It "mutes "`parsedUrl.searchParams`. So using `parsedUrl.searchParams.delete` is more obvious.

patrickvenetz:
> I didn't check if you catched all occurences though.

Double-checked once more and I believe I catched all occurences.

PR #406, opened 2020-04-07
feat: mutations to send test notifications

patte:
- [sendTestPushNotification](http://localhost:5000/graphiql/?query=mutation%20%7BsendTestPushNotification%7D)
- [sendTestNotification](http://localhost:5000/graphiql/?query=mutation%20%7B%0A%20%20sendTestNotification(commentId%3A%20%22837cd81c-b412-4141-a39f-e3589587c883%22%2C%20repoId%3A%20%22republik-dev%2Farticle-scheduled-notification-test%22)%0A%7D%0A)

### DIFF
--- a/packages/discussions/lib/Notifications.js
+++ b/packages/discussions/lib/Notifications.js
@@ -34,6 +34,10 @@ const submitComment = async (comment, discussion, context, testUsers) => {
   const { pgdb, t } = context
   const { id, parentIds, discussionId, userId } = comment
 
+  if (testUsers && !Array.isArray(testUsers)) {
+    throw new Error(t('api/unexpected'))
+  }
+
   const displayAuthor = await getDisplayAuthor(
     comment,
     { portrait: { webp: false } },
--- a/packages/discussions/lib/Notifications.js
+++ b/packages/discussions/lib/Notifications.js
@@ -30,7 +30,7 @@ const getDiscussionUrl = async (discussion, context) => {
   return `${FRONTEND_BASE_URL}${discussion.path}`
 }
 
-const submitComment = async (comment, discussion, context) => {
+const submitComment = async (comment, discussion, context, testUsers) => {
   const { pgdb, t } = context
   const { id, parentIds, discussionId, userId } = comment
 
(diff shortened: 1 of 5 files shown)


packages/discussions/lib/Notifications.js
```diff
@@ -40,7 +40,7 @@ const submitComment = async (comment, discussion, context) => {
     context
   )
 
-  const notifyUsers = await pgdb.query(`
+  const notifyUsers = testUsers || await pgdb.query(`
```
patrickvenetz, line 43: `testUsers` may become truthy but not an array and I think it would be sane to check if `testUsers` is an array or not.

packages/subscriptions/graphql/resolvers/_mutations/sendTestNotification.js
```diff
@@ -0,0 +1,46 @@
+const { ensureUser } = require('@orbiting/backend-modules-auth')
+
+module.exports = async (_, args, context) => {
+  const { submitComment } = require('@orbiting/backend-modules-discussions/lib/Notifications')
+  const { notifyPublish } = require('../../../../../servers/publikator/lib/Notifications')
+
+  const {
+    commentId,
+    docRepoId
```
patrickvenetz, line 9: Just a thought, `repoId` would be enough as we handle no repository IDs but  those of documents.

servers/publikator/lib/Notifications.js
```diff
@@ -29,18 +29,20 @@ const notifyPublish = async (repoId, context) => {
...
-      type: 'Document',
-      id: subscriptionRepoId
-    },
-    context,
-    {
-      onlyEligibles: true
-    }
-  )
-  const subscribers = await Promise.map(
+  const subscriptions = testUsers
```
patrickvenetz, line 32: (2) `testUsers` may come in truthy but not an array and I think it would be sane to check if `testUsers` is an array or not.

servers/publikator/lib/Notifications.js
```diff
@@ -29,18 +29,20 @@ const notifyPublish = async (repoId, context) => {
...
+      {
+        type: 'Document',
+        id: subscriptionRepoId
+      },
+      context,
+      {
+        onlyEligibles: true
+      }
+    )
+  const subscribers = testUsers || await Promise.map(
```
patrickvenetz, line 45: (3) `testUsers` may come in truthy but not an array and I think it would be sane to check if `testUsers` is an array or not.

PR #407, opened 2020-04-08
fix: Remove republik_memberships_count variable from mail templates and merge vars

patrickvenetz:
This Pull Request strips `republik_memberships_count` variable from mail templates as to reduce confusion on number of active members and subscriptions.


patrickvenetz:
giphy

PR #408, opened 2020-04-14
feat(access): Add constraint to check if user ID is in a gsheet list

patrickvenetz:
Constraint checks if granter's user ID is to be found in a `gsheets` record. If not found, constraint will fail. Contraint will hinder display of campaign.

This helps to render campaigns only to users listed in a spreadsheet.

### DIFF
--- a/packages/access/lib/constraints/granterInGsheet.js
+++ b/packages/access/lib/constraints/granterInGsheet.js
@@ -1,0 +1,72 @@
+const debug = require('debug')('access:lib:constraints:granterInGsheet')
+
+/**
+ * Constraint checks if granter's user ID is to be found in a gsheet {settings.name}.
+ * If not found, constraint will fail. Contraint will hinder display of campaign.
+ *
+ * Story: Campaign should only be visible to users listed in a spreadsheet.
+ *
+ * @example: {
+ *  "recipientInGsheet": {
+ *    "name": "accessCampaignSomeGsheet",
+ *    "idenfitifer": "userId",
+ *    "flag": "enabled"
+ *  }
+ * }
+ */
+
+const isGrantable = async (args, context) => {
+  const { granter, settings } = args
+  const { pgdb } = context
+
+  const {
+    name, // gsheets.name field
+    identifier, // row which contains a user ID
+    flag // (optional) row which all enables or disables campaign ("TRUE", "FALSE")
+  } = settings
+
+  if (!name) {
+    debug('settings.name property is falsy')
+  }
+
+  if (!identifier) {
+    debug('settings.identifier property is falsy')
+  }
+
+  const users = await pgdb.public.gsheets.findOneFieldOnly({ name }, 'data')
+
+  if (!users) {
+    debug(`Specified gsheet "${name}" is either missing or emmpty`)
+    return false
+  }
+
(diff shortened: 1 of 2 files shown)


packages/access/lib/constraints/granterInGsheet.js
```diff
@@ -0,0 +1,72 @@
...
+  const { pgdb } = context
+
+  const {
+    name, // gsheets.name field
+    identifier, // row which contains a user ID
+    flag // (optional) row which all enables or disables campaign ("TRUE", "FALSE")
+  } = settings
+
+  if (!name) {
+    debug('settings.name property is falsy')
```
patte, line 29: why not throw or `return false` here?

packages/access/lib/constraints/granterInGsheet.js
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  if (!identifier) {
+    debug('settings.identifier property is falsy')
+  }
+
+  const users = await pgdb.public.gsheets.findOneFieldOnly({ name }, 'data')
+
+  if (!users) {
+    debug(`Specified gsheet "${name}" is either missing or emmpty`)
```
patte, line 39:
```suggestion
    debug(`Specified gsheet "${name}" is either missing or empty`)
```

packages/access/lib/constraints/granterInGsheet.js
```diff
@@ -0,0 +1,72 @@
...
+ * }
+ */
+
+const isGrantable = async (args, context) => {
+  const { granter, settings } = args
+  const { pgdb } = context
+
+  const {
+    name, // gsheets.name field
+    identifier, // row which contains a user ID
```
patte, line 24:
```suggestion
    identifier, // column which contains a user ID
```

packages/access/lib/constraints/granterInGsheet.js
```diff
@@ -0,0 +1,72 @@
...
+    flag // (optional) row which all enables or disables campaign ("TRUE", "FALSE")
```
patte, line 25:
```suggestion
    flag // (optional) column which overwrites grant for user ("TRUE", "FALSE")
```

packages/access/lib/constraints/granterInGsheet.js
patte: `idColumn` or `idColumnName` would have been a better name IMHO.

packages/access/lib/constraints/granterInGsheet.js
```diff
@@ -0,0 +1,72 @@
...
+ *  }
+ * }
+ */
+
+const isGrantable = async (args, context) => {
+  const { granter, settings } = args
+  const { pgdb } = context
+
+  const {
+    name, // gsheets.name field
```
patte, line 23: `sheetName` would have been a better name IMHO.

packages/access/lib/constraints/granterInGsheet.js
patte: `flagColumnName` i'd liked more.

patrickvenetz:
Project aborted most successfully and adding this feature would be unnecessary dead code, hence: Closing PR.

PR #412, opened 2020-04-16
Book structure

annatraussnig:
Support nested parts for series episodes.


packages/documents/graphql/schema-types.js
```diff
@@ -8,12 +8,18 @@ type Series {
...
+  document: Document
+}
+
 type Episode {
   title: String
   label: String
   image: String
   publishDate: DateTime
   document: Document
+  parts: [EpisodePart]
```
patrickvenetz, line 22: If not mistaken, you will return an empty parts array if there are no parts to it, right?

If so, schema is tighter with requiring a set to be returned `[...]!` (and returned entities are of `EpisodePart`, always, so `EpisodePart!`)

```suggestion
  parts: [EpisodePart!]!
```

patrickvenetz:
Closing, while "labour"-intensive work, not about to be merge in foreseeable future.

PR #418, opened 2020-04-28
feat: featured comments

patte:
### GraphiQL
- [mutation](http://localhost:5000/graphiql/?query=mutation%20%7B%0A%20%20featureComment(id%3A%20%224ccf44b9-a660-49bd-a0f6-4c5ce455d2c3%22%2C%20content%3A%20%22test%20test%22)%20%7B%0A%20%20%20%20id%0A%20%20%20%20featuredAt%0A%20%20%20%20featuredText%0A%20%20%20%20featuredContent%0A%20%20%7D%0A%7D%0A)
- [query](http://localhost:5000/graphiql/?query=%7B%0A%20%20comments(orderBy%3A%20FEATURED_AT%2C%20orderDirection%3A%20DESC%2C%20first%3A%2015%2C%20featured%3A%20true)%20%7B%0A%20%20%20%20totalCount%0A%20%20%20%20nodes%20%7B%0A%20%20%20%20%20%20id%0A%20%20%20%20%20%20text%0A%20%20%20%20%20%20featuredAt%0A%20%20%20%20%20%20featuredText%0A%20%20%20%20%20%20featuredContent%0A%20%20%20%20%7D%0A%20%20%7D%0A%7D%0A)


packages/discussions/graphql/resolvers/Comment.js
```diff
@@ -138,6 +138,16 @@ module.exports = {
   text: (comment, args, context) =>
     textForComment(comment, false, context),
 
+  featuredContent: ({ published, adminUnpublished, featuredAt, featuredContent }) =>
```
tpreusse, line 141: Currently all featured comment text is plain text right now like `comment.preview`. I would prefer to remove the `featuredContent` resolver and only provide `featuredText` for now.

packages/discussions/graphql/resolvers/_mutations/featureComment.js
```diff
@@ -0,0 +1,45 @@
...
+  const comment = await loaders.Comment.byId.load(id)
+  if (!comment) {
+    throw new Error(t('api/comment/404'))
+  }
+
+  const newComment = await pgdb.public.comments.updateAndGetOne(
+    { id },
+    {
+      featuredAt: new Date(),
+      featuredContent: content
```
tpreusse, line 27: Currently it's possible to pass an empty string and then receive `null` for `featuredText` but a `featuredAt` date. This would currently create problems for the front teaser which always needs a `featuredText`.

Currently there is also noway of «unfeaturing». This might be needed because of accidental features or someone might change their mind.

My suggestion: Let's make `content` arg of the mutation nullable in the schema and null `featuredAt` and `featuredContent` when `content` is falsy. The alternative would be to ensure `content` is not empty and add an unfeature mutation.

packages/discussions/graphql/resolvers/_mutations/featureComment.js
```diff
@@ -0,0 +1,45 @@
...
+      comment: {
+        mutation: 'UPDATED',
+        node: newComment
+      }
+    }),
+  ])
+    .catch(e => {
+      console.error(e)
+    })
+
```
tpreusse, line 43: Let's add a slack notifications which includes the context user. Just to control against mysterious features.

PR #420, opened 2020-05-04
feat(maillog): Add errornous record filter for mail log records

patrickvenetz:
This Pull Request adds `filters: { errornous: TRUE|FALSE }` to mail log queries.

```gql
{
  mailLog(filters: {errornous: true}) {
    totalCount
    nodes {
      email
      error
    }
  }
  me {
    mailLog(filters: {errornous: false}) {
      totalCount
      nodes {
        error
      }
    }
  }
}
```

Code is slightly refactored and extended, moving pagination (including filtering) into a lib-file as well as record props conditionals (error, status, template).

### DIFF
--- a/packages/maillog/graphql/schema-types.js
+++ b/packages/maillog/graphql/schema-types.js
@@ -1,7 +1,7 @@
 module.exports = `
 
 input MaiLogFiltersInput {
-  errornous: Boolean
+  hasError: Boolean
 }
 
 type MailLogConnection {
--- a/packages/maillog/graphql/schema-types.js
+++ b/packages/maillog/graphql/schema-types.js
@@ -1,5 +1,9 @@
 module.exports = `
 
+input MaiLogFiltersInput {
+  errornous: Boolean
+}
+
 type MailLogConnection {
   totalCount: Int!
   pageInfo: MailLogPageInfo!
@@ -38,6 +42,7 @@ extend type User {
     last: Int
     before: String
     after: String
+    filters: MaiLogFiltersInput
   ): MailLogConnection
 }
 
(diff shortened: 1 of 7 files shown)


packages/maillog/graphql/schema-types.js
```diff
@@ -1,5 +1,9 @@
 module.exports = `
 
+input MaiLogFiltersInput {
+  errornous: Boolean
```
patte, line 4: typo, the thing is spelled: `erroneous`. Why not just `error`?

packages/maillog/lib/paginate.js
```diff
@@ -0,0 +1,29 @@
...
+  return stack
+}
+
+const paginate = (records, args) => paginator(
+  { first: 100, ...args },
+  a => a,
+  ({ filters = {} }) => {
+    const filterFns = getFilterFns(filters)
+
+    return records.filter(record => filterFns.every(filterFn => filterFn(record)))
```
patte, line 23:
```suggestion
    return filterFns.length
      ? records.filter(record => filterFns.every(filterFn => filterFn(record)))
      : records
```

packages/maillog/graphql/schema-types.js
patrickvenetz: `error` won't indicate a `Boolean` value, considering `MailLogRecord.error` is a `String`.

Settle for `hasError`!

packages/maillog/lib/paginate.js
patrickvenetz: Thanks!

patrickvenetz:
@patte Please revisit.

PR #430, opened 2020-05-18
refactor: get rid of passport, decrease session prune interval, prime user dataloader with me

patte:


patte:
error of not [removing countryFlag completely](https://github.com/orbiting/backends/pull/430/files#diff-9e86d4abec0d4d1c6efbc5764dcef24dR7) removed with: https://github.com/orbiting/backends/commit/539d581df8dcd942928e12ba114dee57cadfa835

PR #434, opened 2020-05-21
Engagement Metrics w/ Scheduler

patrickvenetz:
This Pull Request is updating data for engagement metrics with a scheduler. It is a variation on https://github.com/orbiting/backends/pull/433.

While developing, I ran into a pickle, wondering if updating pre-populated data via a cronjob is a better, leaner solution:

- Updating engagement data is not required within minutes
- Scheduler consumes credits (memory, CPU) on each instance


patrickvenetz:
Discarded.

PR #436, opened 2020-05-24
feat(publikator): Configure MailChimp campaign segment, envelope via configs ENV var

patrickvenetz:
Use `MAILCHIMP_CAMPAIGN_CONFIGS` to set configurations, containing array with objects.

Each object may contain:
- key, currently format.meta.repoId)
- list_id, MailChimp List ID
- saved_segment_id, MailChimp Segment ID
- to_name, MailChimp Merge Vars as To: name
- from_name, sender From: name
- reply_to, sender email

Publish mutation checks if key matches format.meta.repoId and merges configuration.

This includes fixes:
* Neater error handling if MailChimp fetch request fails or returns errorneous status
* Updated subject line or title will be pushed to MailChimp which it did before only when creating
* From: defaults to DEFAULT_MAIL_FROM_NAME and DEFAULT_MAIL_ADDRESS

Example value in `MAILCHIMP_CAMPAIGN_CONFIGS`

```json
[
  {
    "key": "republik-dev/format-newsletter-um-7",
    "from_name": "Republik (Staging)",
    "reply_to": "kontakt@republik.love",
    "list_id": "xxx",
    "saved_segment_id": 1234
  },
  {
    "key": "republik-dev/format-wochenende-newsletter",
    "from_name": "Republik (Staging)",
    "reply_to": "kontakt@republik.love",
    "list_id": "xxx",
    "saved_segment_id": 2345
  },
  {
    "key": "republik-dev/format-covid-19-uhr-newsletter",
    "from_name": "Republik (Staging)",
    "reply_to": "covid19@republik.love",
    "list_id": "95ae41fe21",
    "saved_segment_id": 3456
  }
]
```

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -132,6 +132,7 @@ DEFAULT_MAIL_FROM_NAME="Republik"
 #  - republik to set interests based on belongings of a user
 #MAILCHIMP_URL=https://us9.api.mailchimp.com
 #MAILCHIMP_API_KEY=
+#MAILCHIMP_CAMPAIGN_CONFIGS='[{"key":"org/some-format","from_name":"Republik"}]'
 
 #############
 # assets
--- a/.env.example
+++ b/.env.example
@@ -132,7 +132,6 @@ DEFAULT_MAIL_FROM_NAME="Republik"
 #  - republik to set interests based on belongings of a user
 #MAILCHIMP_URL=https://us9.api.mailchimp.com
 #MAILCHIMP_API_KEY=
-#MAILCHIMP_CAMPAIGN_CONFIGS='[{"key":"org/some-format","from_name":"Republik"}]'
 
 #############
 # assets
--- a/packages/translate/translations.json
+++ b/packages/translate/translations.json
@@ -1,5 +1,5 @@
 {
-  "updated": "2020-03-23T20:34:06.306Z",
+  "updated": "2020-05-24T13:10:16.770Z",
   "title": "live",
   "data": [
     {
@@ -1693,6 +1693,18 @@
     {
       "key": "api/publish/disabled",
       "value": "Artikel können im Moment weder veröffentlicht noch zurückgezogen werden. Bitte versuche es in ein paar Minuten noch einmal, ansonsten kontaktiere bitte die IT."
+    },
+    {
+      "key": "api/publish/error/createCampaign",
+      "value": "Kampagne auf MailChimp konnte nicht angelegt werden."
+    },
+    {
+      "key": "api/publish/error/updateCampaign",
+      "value": "Kampagne auf MailChimp konnte nicht aktualisiert werden."
+    },
+    {
+      "key": "api/publish/error/updateCampaignContent",
+      "value": "Kampagne auf MailChimp konnte nicht aktualisiert werden."
     },
     {
       "key": "api/archive/error/published",
--- a/servers/publikator/.env.example
+++ b/servers/publikator/.env.example
@@ -6,7 +6,19 @@ CORS_WHITELIST_URL=http://localhost:3005
 # analytics with apollo's engine
 #ENGINE_API_KEY=
 
-# default settings for mailchimp campaign creation
+# MAILCHIMP_CAMPAIGN_CONFIGS is a way to default MailChimp settings when pushed
+# to MailChimp like a segment or sender address.
+#
+# It is expected to be Array<Object> and each Object may contain:
+#
+# {
+#   "key": "<Repo ID to a format>",
+#   "from_name": "Republik",
+#   "reply_to": "<Sender email address>",
+#   "list_id": "<MailChimp List ID>",
+#   "saved_segment_id": <MailChimp Segment ID in List ID>
+# }
+#
 #MAILCHIMP_CAMPAIGN_CONFIGS='[{"key":"org/some-format","from_name":"Republik"}]'
 
 # filter repos by their names
--- a/servers/publikator/.env.example
+++ b/servers/publikator/.env.example
@@ -7,8 +7,7 @@ CORS_WHITELIST_URL=http://localhost:3005
 #ENGINE_API_KEY=
 
 # default settings for mailchimp campaign creation
-MAILCHIMP_FROM_NAME=Republik
-MAILCHIMP_REPLY_TO=kontakt@republik.ch
+#MAILCHIMP_CAMPAIGN_CONFIGS='[{"key":"org/some-format","from_name":"Republik"}]'
 
 # filter repos by their names
 #REPOS_NAME_FILTER=newsletter-,article-
--- a/servers/publikator/lib/mailchimp/createCampaign.js
+++ b/servers/publikator/lib/mailchimp/createCampaign.js
@@ -1,27 +1,18 @@
 const fetch = require('isomorphic-unfetch')
 
-const {
-  MAILCHIMP_URL,
-  MAILCHIMP_API_KEY,
-  MAILCHIMP_FROM_NAME,
-  MAILCHIMP_REPLY_TO
-} = process.env
+const { MAILCHIMP_URL, MAILCHIMP_API_KEY } = process.env
 
-module.exports = async ({ title, subject }) => {
+module.exports = async () => {
   return fetch(`${MAILCHIMP_URL}/campaigns`, {
     method: 'POST',
     headers: {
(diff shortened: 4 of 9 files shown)


servers/publikator/.env.example
```diff
@@ -7,8 +7,7 @@ CORS_WHITELIST_URL=http://localhost:3005
 #ENGINE_API_KEY=
 
 # default settings for mailchimp campaign creation
-MAILCHIMP_FROM_NAME=Republik
-MAILCHIMP_REPLY_TO=kontakt@republik.ch
+#MAILCHIMP_CAMPAIGN_CONFIGS='[{"key":"org/some-format","from_name":"Republik"}]'
```
patte, line 10: ENVs are confusing already, so don't repeat the config from `.env.example` here. I'm for the option of only setting it in `.env.example`

.env.example
```diff
@@ -132,6 +132,7 @@ DEFAULT_MAIL_FROM_NAME="Republik"
 #  - republik to set interests based on belongings of a user
 #MAILCHIMP_URL=https://us9.api.mailchimp.com
 #MAILCHIMP_API_KEY=
+#MAILCHIMP_CAMPAIGN_CONFIGS='[{"key":"org/some-format","from_name":"Republik"}]'
```
patte, line 135: Describe the schema and possible keys here or link to this PR's description.

packages/translate/translations.json
```diff
@@ -1694,6 +1694,18 @@
...
+      "key": "api/publish/error/createCampaign",
+      "value": "Kampagne auf MailChimp konnte nicht angelegt werden."
+    },
+    {
+      "key": "api/publish/error/updateCampaign",
+      "value": "Kampagne auf MailChimp konnte nicht aktualisiert werden."
+    },
+    {
+      "key": "api/publish/error/updateCampaignContent",
+      "value": "Kampagne auf MailChimp konnte nicht aktualisiert werden."
```
patte, line 1707:
```suggestion
      "value": "Kampagne auf MailChimp konnte nicht aktualisiert werden. (updateCampaignContent)"
```

packages/translate/translations.json
```diff
@@ -1694,6 +1694,18 @@
       "key": "api/publish/disabled",
       "value": "Artikel können im Moment weder veröffentlicht noch zurückgezogen werden. Bitte versuche es in ein paar Minuten noch einmal, ansonsten kontaktiere bitte die IT."
     },
+    {
+      "key": "api/publish/error/createCampaign",
+      "value": "Kampagne auf MailChimp konnte nicht angelegt werden."
+    },
+    {
+      "key": "api/publish/error/updateCampaign",
+      "value": "Kampagne auf MailChimp konnte nicht aktualisiert werden."
```
patte, line 1703:
```suggestion
      "value": "Kampagne auf MailChimp konnte nicht aktualisiert werden. (updateCampaign)"
```

servers/publikator/lib/mailchimp/updateCampaign.js
```diff
@@ -0,0 +1,54 @@
...
+    ...campaignConfig
+  }
+
+  const body = {
+    recipients: {
+      ...config.list_id && { list_id: config.list_id },
+      segment_opts: {
+        ...config.saved_segment_id && { saved_segment_id: Number(config.saved_segment_id) }
+      }
+    },
```
patte, line 28: What if the config doesn't specify `list_id` or `saved_segment_id`? Does sending mailchimp an empty `recipients` key reset what was maybe set manually on mailchimp?

servers/publikator/lib/mailchimp/createCampaign.js
```diff
@@ -30,6 +21,10 @@ module.exports = async ({ title, subject }) => {
...
     })
   })
-    .then(response => response.json())
-    .catch(error => console.error('updateMailchimp failed', { error }))
+    .then(response => {
+      if (!response.ok) {
+        throw Error(response.statusText)
+      }
+      return response
+    })
```
patte, line 29: There was [a check](https://github.com/orbiting/backends/pull/436/files#diff-6e358742c5bc46972ccb65ea7e02ad7fL277) for `status !== 'save'` before, would it not be worth keeping that to ensure mailchimp did really save not just the request didn't fail?

servers/publikator/lib/mailchimp/updateCampaignContent.js
```diff
@@ -10,12 +10,16 @@ module.exports = async ({ campaignId, html }) => {
...
     })
   })
-    .then(response => response.json())
-    .catch(error => console.error('updateMailchimp failed', { error }))
+    .then(response => {
+      if (!response.ok) {
+        throw Error(response.statusText)
+      }
+      return response
+    })
```
patte, line 24: Before, the response was returned as json, and I think that's nice of a lib to do instead of returning a fetch-respose. Also there was [a check](https://github.com/orbiting/backends/pull/436/files#diff-6e358742c5bc46972ccb65ea7e02ad7fL457) for `response.status` to be missing before. Maybe it would be good to keep this check.

servers/publikator/lib/mailchimp/updateCampaign.js
```diff
@@ -0,0 +1,54 @@
...
+  DEFAULT_MAIL_FROM_ADDRESS
+} = process.env
+
+const mailchimpCampaignConfigs = MAILCHIMP_CAMPAIGN_CONFIGS
+  ? JSON.parse(MAILCHIMP_CAMPAIGN_CONFIGS)
+  : []
+
+module.exports = async ({ campaignId, campaignConfig = {} }) => {
+  const config = {
+    ...mailchimpCampaignConfigs.find(c => c.key === campaignConfig.key),
```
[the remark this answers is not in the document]
patrickvenetz: Because spreading does not return anything to merge nor does it throw an error, handling seems unnecessary.

```
> { ...[{a: 1}, {a: 2}].find(o => o.a === 3) }
{}
> { ...null }
{}
> { ...undefined }
{}
> { ...false }
{}
```

servers/publikator/lib/mailchimp/createCampaign.js
patrickvenetz: MailChimp will either return a response which is thrown as an error if `response.ok` is falsy ([status code outside 200-299 range](https://developer.mozilla.org/en-US/docs/Web/API/Response/ok)), or fetch itself will throw an error.

Thus sticking to current setup, believing it handles MailChimp API (and fetch) rather well.

servers/publikator/lib/mailchimp/updateCampaignContent.js
patrickvenetz: `response.status` was good, `response.ok` fits even better in case of MailChimp API.

Their API really sticks to (resource) status codes. Whatever you `POST`, `PATCH`, `PUT` you will either result in a good status code (200-299, thus `response.ok` becomes truthy) or bad one.

servers/publikator/lib/mailchimp/updateCampaign.js
patrickvenetz: Good question, really. But no, it will not overwrite anything if either prop is not posted.

(If `list_id` and `saved_segment_id` is set, this _will_ overwrite whatever one might have changed on MailChimp (same with content). This is intended, declaring Publikator as the single source of truth here.)

packages/translate/translations.json
patrickvenetz: This error message (and `updateCampaign`) are intended for regular users, indicating the realm of the error (campaign update failed). While this is posted via API, [we will log an error with stack trace](https://github.com/orbiting/backends/pull/436/files#diff-6e358742c5bc46972ccb65ea7e02ad7fR478) and enabling us to debug.

servers/publikator/.env.example
patrickvenetz: So true. Changing.

.env.example
patrickvenetz: Added a more verbose comment.

patrickvenetz:
About how long MailChimp API requests take, tracked w/ `console.time()`:

```
createCampaign: 889.057ms

updateCampaign: 271.935ms
updateCampaign: 209.5ms
updateCampaign: 266.305ms

updateCampaignContent: 1.197s
updateCampaignContent: 1.039s
updateCampaignContent: 1.371s
```

PR #447, opened 2020-06-14
Black Lives Matter

patte:
replace blacklist -> denylist, whitelist -> allowlist

affected ENVS:
```
lambda:
URL_WHITELIST

backends:
CORS_WHITELIST_URL
AWS_BUCKET_WHITELIST
```


patrickvenetz:
OSS seems to gravitate more towards deny list and allow list, see https://www.google.com/search?q=allowlist+denylist

PR #449, opened 2020-06-17
feat(republik): Accept optional payload data in mutation submitPledge

patrickvenetz:


servers/republik/migrations/sqls/20200617122258-submit-pledge-payload-down.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE "public"."pledges" DROP COLUMN IF EXISTS "payload" jsonb ;
```
patte, line 1:
```suggestion
ALTER TABLE "public"."pledges" DROP COLUMN IF EXISTS "payload" jsonb;
```

servers/republik/migrations/sqls/20200617122258-submit-pledge-payload-up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE "public"."pledges" ADD COLUMN IF NOT EXISTS "payload" jsonb ;
```
patte, line 1:
```suggestion
ALTER TABLE "public"."pledges" ADD COLUMN IF NOT EXISTS "payload" jsonb;
```

patrickvenetz:
I feel the urge.

Schema somewhat ensures format, that would be JSON. Restrictions come naturally with either database limit or GraphQL request body size. User generated content might end up in database but likely not disturbing it.

Due to the juvenile feature I'd argue to not restrict it with a detailed schema just yet and accept risk of having unwanted but not harmful data in payload.

PR #453, opened 2020-06-25
Subscribe authors / users (comments)

patte:
basically a resurrection and modernization of 5a7976c

featuring optional chaining and nullish coalescing in one and the same [line](https://github.com/orbiting/backends/pull/453/files#diff-09ab43947bba51ac919009fbf7776378R171) :tada: 

before deploy:
- [ ] set all existing subscriptions to users (originating for the politics tindr) to `active=false`.


patte:
open question: if a user set the notificationOption of a discussion to NONE 'Keine', then he also doesn't get notified if a subscribed users comments in that discussion. I guess that's what we want right? @tpreusse @patrickvenetz

PR #456, opened 2020-06-29
refactor(republik): Use MembershipStats.evolution to provide MembershipStats.count

patrickvenetz:
This Pull Requests changes calcuation of `MembershipStats.count`: It adds `active` and `overdue` memberships from current `MembershipStats.evolution` bucket.

servers/republik/lib/MembershipStats/evolution.js is extended with `sumBucketProps`. It returns a sum of properites `props` in bucket `key` from cache.

Example:

```js
sumBucketProps(context, '2020-05', { add: ['active'], subtract: ['cancelled'] }
```

… would return bucket.active minus bucket.cancelled of bucket with key "2020-05".

`sumBucketProps` defaults to current bucket (current month) and returns number of "active" plus "overdue" numbers.

First application is in `MembershipStats.count` (implemented), further ones in MailChimp Merge Vars (not implemented).


servers/republik/graphql/resolvers/MembershipStats/count.js
```diff
@@ -1,14 +1,9 @@
...
-  `, {
-    excludeUserId: PARKING_USER_ID
-  })
+const { sumBucketProps } = require('../../../lib/MembershipStats/evolution')
+
+module.exports = (_, args, context) => sumBucketProps(
+  context,
+  moment().format('YYYY-MM'),
+  { add: ['active', 'overdue'] }
+)
```
patte, line 9:
```suggestion
) || -1
```
as `sumBucketProps` might return null

servers/republik/lib/MembershipStats/evolution.js
```diff
@@ -219,7 +219,53 @@ const populate = async (context, resultFn) => {
...
+  const bucket = result.find(bucket => bucket.key === key)
+
+  if (!bucket) {
+    return null
+  }
+
+  const { add = ['active', 'overdue'], subtract = [] } = props
+
+  const sumAdd = add.reduce((prev, curr) => prev + bucket[curr], 0)
+  const sumSubtract = subtract.reduce((prev, curr) => prev + bucket[curr], 0)
```
patte, line 262: mybe it would be helfull to log an error if a `add` or `subtract` key can't be found in `bucket`? Or sanitize throw if an invalid key is specified in `props`?

servers/republik/lib/MembershipStats/evolution.js
```diff
@@ -219,7 +219,53 @@ const populate = async (context, resultFn) => {
...
+  key = moment().format('YYYY-MM'),
+  props
+) => {
+  // Fetch pre-populated data
+  const data = await createCache(context).get()
+
+  // Return null in case pre-populated data is not available
+  if (!data) {
+    return null
+  }
```
patte, line 247:
```suggestion
  if (!data) {
    console.error('pre-populated data is not available')
    return null
  }
```

servers/republik/lib/MembershipStats/evolution.js
```diff
@@ -219,7 +219,53 @@ const populate = async (context, resultFn) => {
...
+
+  // Retrieve pre-populated result from data.
+  const { result = [] } = data
+
+  // Find desired bucket
+  const bucket = result.find(bucket => bucket.key === key)
+
+  if (!bucket) {
+    return null
```
patte, line 256:
```suggestion
    console.error(`bucket ${key} not found`)
    return null
```

servers/republik/graphql/resolvers/MembershipStats/count.js
patrickvenetz: Not falling back should throw a GraphQL error if `sumBucketProps` returns `null`, intentionally.

You sort of prefer to always return a value, even a faulty one?

servers/republik/lib/MembershipStats/evolution.js
patrickvenetz: Yes, very good.

servers/republik/lib/MembershipStats/evolution.js
patrickvenetz: Logging very good, using WARN level instead. Keeping ERROR level for those backends is throwing, too.

patrickvenetz:
@patte Please have another peak.

Change of error strategy: Throwing in lib, catching them in resolver and throwing a generic error, but still logging.

PR #457, opened 2020-07-01
Mutation: Append Period to Membership

caillou:


caillou:
@patrickvenetz this PR implements an API in order to add a period to any membership.

My intuition tells me, this should not be implemented on a monthly membership. Am I wrong?

patrickvenetz:
As state in stand-up this very early morning: Your intuition is spot on.

servers/republik/modules/crowdfundings/graphql/resolvers/_mutations/appendPeriodToMembership.js
```diff
@@ -0,0 +1,66 @@
...
+    return now
+  }
+
+  return lastEndDate
+}
+
+function getEndDate (startDate, duration, durationUnit) {
+  const endDate = moment(startDate).add(duration, durationUnit).toDate()
+  if (endDate.toString() === startDate.toString()) {
+    throw new Error(`${startDate}, ${duration}, ${durationUnit}`)
```
patrickvenetz, line 63: I find this error message rather unworthy of its cause. Maybe slip in a human readable hint?

servers/republik/modules/crowdfundings/graphql/schema-types.js
```diff
@@ -165,6 +165,7 @@ type Membership {
   graceEndDate: DateTime
   overdue: Boolean!
   cancellations: [Cancellation!]!
+  canManuallyAddPeriod: Boolean!
```
patrickvenetz, line 168: Love that flag.

I urge you to rename as close as possible to its sibling mutation as to have one mental hoop less to jump through.

So, here is a list of these:

- `canAddPeriod` (prop), `addPeriod` (mutation)
- `canAppendPeriod`, `appendPeriod`  ⭐
- `canManuallyAddPeriod`, `addPeriod`
- `canManuallyAppendPeriod`, `appendPeriod`
- `canAddMembershipPeriod`, `addMembershipPeriod`
- `canAppendMembershipPeriod`, `appendMembershipPeriod` ⭐

Those I would pick myself, I marked with a  ⭐.

Here some reasoning around `manual` and `toMembership`, or why I favor dropping it:

- There is no other entity you can add a period to.
- All mutation are somewhat manual.
- Using entity name itself (`MembershipPeriod`) is good. One hoop less.

servers/republik/modules/crowdfundings/graphql/resolvers/Membership.js
```diff
@@ -178,6 +178,16 @@ const membershipResolver = {
     const type = await membershipResolver.type(membership, null, context)
 
     return type.interval === 'year'
+  },
+
+  async canManuallyAddPeriod ({ subscriptionId: stripeSubscriptionId, active }, x, { pgdb }) {
```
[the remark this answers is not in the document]
patrickvenetz: There is probably little need to alias `subscriptionId` to `stripeSubscriptonId`.

servers/republik/__tests__/crowdfundings/integration.jest.js
```diff
@@ -487,3 +487,154 @@ describe('cancelPledge', () => {
...
+    })
+
+    expect(result.errors).toBeTruthy()
+    expect(result.errors.length).toEqual(1)
+    expect(result.errors[0].message).toEqual(
+      'Die durationShouldBePositive Schnittstelle erwartet ein positives `duration` Feld. Es wurde folgender Wert übergeben: -99.'
+    )
+  })
+
+  test('appends a periods', async () => {
```
patrickvenetz, line 587:
```suggestion
  test('appends periods', async () => {
```

servers/republik/modules/crowdfundings/graphql/resolvers/_mutations/appendPeriodToMembership.js
```diff
@@ -0,0 +1,66 @@
...
+    await transaction.transactionRollback()
+    console.error('appendPeriodToMembership', e, { req: req._log() })
+    throw e
+  }
+}
+
+async function getBeginDate (transaction, id) {
+  const result = await transaction.public.membershipPeriods.findFirst({
+    membershipId: id
+  }, { fields: '"endDate"', orderBy: ['endDate desc'] })
```
patrickvenetz, line 43: Wondering if double-quoting of `"endDate"` is necessary because pogi does not, but does in `orderBy:`?

servers/republik/modules/crowdfundings/graphql/resolvers/_mutations/appendPeriodToMembership.js
```diff
@@ -0,0 +1,66 @@
...
+const moment = require('moment')
+
+module.exports = async (_, { id, duration, durationUnit }, context) => {
+  const { user: me, t, req } = context
+  Roles.ensureUserHasRole(me, 'supporter')
+
+  const transaction = await context.pgdb.transactionBegin()
+
+  try {
+    const membership = await transaction.public.memberships.findOne({ id })
```
patrickvenetz, line 11: Considering this mutation can run at any time, without knowledge of actions before, it should test some as `canManuallyAddPeriod`.

servers/republik/modules/crowdfundings/graphql/resolvers/_mutations/appendPeriodToMembership.js
```diff
@@ -0,0 +1,66 @@
...
+    const endDate = getEndDate(beginDate, duration, durationUnit)
+
+    await transaction.public.membershipPeriods.insert({
+      membershipId: id,
+      beginDate,
+      endDate,
+      kind: 'ADMIN'
+    })
+
+    const updatedMembership = await transaction.public.memberships.findOne({ id })
```
patrickvenetz, line 30: You did not alter membership (but its periods) and there is no actual need to refetch that membership. Returning initial `membership` is enough.

(A plausible `Membership.periods` query will be resolved after this mutation fetch a new period, too.)

servers/republik/modules/crowdfundings/graphql/resolvers/Membership.js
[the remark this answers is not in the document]
caillou: @patrickvenetz with regard to the `stripeSubscriptionId`: The idea was to put an emphasis on the fact, that this is not the `id` of the `subscriptions` table, which I initially thought and lost time with. Obviously, this renaming should happen on the DB field, not here.

Given this argument, would you still not using the alias?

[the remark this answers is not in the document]
patrickvenetz: Haven't even thought of it, and find that a rather compelling argument. Especially since you fell for it.

(I'd probably still not use the alias, arguing somewhat on the scope in which this resolver is to be found and what it handles, both membershipy stuff. Also: worked-too-long-with-that-subscription-ID-bias.)

Not taking a stand here, `stripeSubscriptionId` is good to and it reflects very much what it is.

servers/republik/modules/crowdfundings/graphql/resolvers/_mutations/appendPeriodToMembership.js
caillou: This makes a whole lot of sense and was my intuition on the previous PR.

As discussed with @patte, there could be some Postgres functions or trigges that actually modify the membership table. Hence it might be a good idea to always re-fetch the object. Is this ridiculous?

servers/republik/modules/crowdfundings/graphql/resolvers/_mutations/appendPeriodToMembership.js
caillou: How could I have forgotten that?!?

PR #459, opened 2020-07-03
tests: speed up, less debug (elastic)

patte:


packages/search/lib/dropElasticsearch.js
```diff
@@ -1,9 +1,13 @@
+let debug = console.log
 const Elasticsearch = require('@orbiting/backend-modules-base/lib/Elasticsearch')
 
-const drop = async (prefix) => {
+const drop = async (prefix, doDebug) => {
   if (!prefix) {
     throw new Error("can't drop, prefix not specified")
   }
+  if (doDebug !== undefined && !doDebug) {
```
caillou, line 8:
```suggestion
const drop = async (prefix, doDebug = true) => {
  if (!prefix) {
    throw new Error("can't drop, prefix not specified")
  }
  if (!doDebug) {
```

Gives me less to think ;)

packages/search/lib/dropElasticsearch.js
```diff
@@ -1,9 +1,13 @@
...
+    debug = identity => identity
```
caillou, line 9: If you call this function as follows, the result will not be as expected:

```
debug(prefix1, false)
debug(prefix2)
```

Maybe we could just do the following:

```
const debug = doDebug === false ? i => i : console.log
```

caillou: also, why does it need the identy function? `console.log` returns `undefined`.

packages/search/lib/pullElasticsearch.js
```diff
@@ -15,8 +15,13 @@ module.exports = async ({
   indices: indicesFilter = mappings.list.map(({ name }) => name),
   switch: doSwitch = true,
   inserts: doInserts = true,
-  flush: doFlush = false
+  flush: doFlush = false,
+  wait: doWait = true,
+  debug: doDebug
 }) => {
+  if (doDebug !== undefined && !doDebug) {
```
caillou, line 22: Same comment as in `packages/search/lib/dropElasticsearch.js`.

packages/search/lib/pullElasticsearch.js
```diff
@@ -112,8 +119,10 @@ module.exports = async ({
         redis
       })
 
-      debug('waiting grace period', { writeAlias, index })
-      await timeout(1000 * 5)
+      if (doWait) {
```
caillou, line 122: Lines 107 - 110 look like copy paste. Could you generalize it?

Something like `await inserts.dict[name].after({}).then(logAndReturn)`

At the same time, you could choose a function name that might suggest why this is sometimes needed? Or we could pass a variable with a name that tells us why we need to wait, instead of what we do?

e.g. `foo({ensurePropagation: true})` statt `foo({doWait: true})`?

packages/test/instance/index.js
```diff
@@ -68,9 +68,11 @@ const init = async ({ serverName, publicationScheduler, searchNotifyListener = n
   // create ES indices
   const pullElasticsearch = require('@orbiting/backend-modules-search/lib/pullElasticsearch')
   const dropElasticsearch = require('@orbiting/backend-modules-search/lib/dropElasticsearch')
-  await dropElasticsearch(esPrefix)
+  await dropElasticsearch(esPrefix, false)
```
caillou, line 71: These would speak more to s.o. who reads the code in 2 weeks:

```
dropElasticsearch({prefix: esPrefix, debug: false}) // probably too much work
dropElasticsearch(esPrefix, {debug: false})
```

caillou:
General question: how much do we gain from the `doDebug: false` as opposed to the `wait: false`?

packages/search/lib/pullElasticsearch.js
patte: I'd rather not refactor the 2 copy/pasted lines. The style of logging in this file could be better, but that's out of scope for my PR.

packages/search/lib/dropElasticsearch.js
patte: re identity, do you have another suggestion? Would you like `a => false` better?

caillou: `() => {}`, or `function () {}`

PR #466, opened 2020-07-09
fix: upsertDiscussion only at time of publish

### LINKED ISSUE #458: Discussion path changes when discussion document is only scheduled
## Bug Report

URL of an already published discussion is updated when only scheduling (discussion) document.

**Steps to Reproduce**

1. Create a discussion document
2. Publish discussion document
3. Change publication date
4. Schedule publication of discussion document

**Actual Results**

`discussions.path` is updated upon scheduling.

**Expected Results**

`discussions.path` should only be updated while publishing.

patte:
fixes #458

### DIFF
--- a/packages/discussions/lib/Discussion.js
+++ b/packages/discussions/lib/Discussion.js
@@ -33,6 +33,7 @@ const upsert = async (id, settings = {}, { pgdb, loaders }, legacyDiscussionId) 
       (settings.maxLength && settings.maxLength !== discussion.maxLength) ||
       (settings.minInterval && settings.minInterval !== discussion.minInterval) ||
       (settings.anonymity && settings.anonymity !== discussion.anonymity) ||
+      (settings.path && settings.path !== discussion.path) ||
       (
         settings.closed !== undefined &&
         settings.closed !== null &&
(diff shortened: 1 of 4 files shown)


patte:
> Did you QA that?

Yes I did. But maybe you could test it out yourself tomorrow aswell?
Putting `{ range: { 'meta.scheduledAt': { lte: new Date('2020-07-12:18:40:00.000+02') } } }` into [PublicationScheduler.js](https://github.com/orbiting/backends/blob/master/servers/publikator/lib/PublicationScheduler.js#L33) helps to simulate another time.

What was the unexpected end result of this bug again? For me it didn't do anything funny, even when `discussion.path` was not correct. I could still open the article and everything.

patrickvenetz:
This bug caused already published "discussion" documents to point a non existing path.

You published a "discussion" on May 15, 2020, thus is `/2020/05/15/slug/diskussion`.

You'll update publication date afterwards to May 20, 2020 (an update!) and only schedule it.

Scheduling will set new path in database table `discussions` already to new date, resulting in `/2020/05/20/slug/diskussion`, but won't create a redirection record (15th to 20th). Hereafter `/2020/05/15/slug/diskussion` can't be found anymore.

PR #471, opened 2020-07-14
feat: github downtime resilience

patte:
make repo overview, commits tree and milestones available even if github is unavailable

new redis cache, which is only used if github doesn't answer
- commits (heads)
- annotatedTags

use repo cache for single repo lookup (repo query)

throw error if github is not accessible and cache not available

### DIFF
--- a/servers/publikator/lib/github/index.js
+++ b/servers/publikator/lib/github/index.js
@@ -365,7 +365,7 @@ module.exports = {
       .then(tags =>
         uniqWith(tags, (a, b) => a.name === b.name)
       )
-      .catch(e => {})
+      .catch(e => {console.error(e)})
 
     // github downtime resilience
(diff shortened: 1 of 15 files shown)


servers/publikator/lib/github/index.js
```diff
@@ -366,17 +365,26 @@ module.exports = {
       .then(tags =>
         uniqWith(tags, (a, b) => a.name === b.name)
       )
+      .catch(e => {})
```
tpreusse, line 368: shouldn't we log this?

tpreusse (commented):
I have one question re error logging but otherwise I think we should just roll it out. Anna tested it locally and worked.

servers/publikator/lib/github/index.js
patte: I added logging there. It's going to fire every time github is down, but I guess that's ok.

PR #475, opened 2020-07-22
Changes serving Admin-Tool

patrickvenetz:
- Post slack message when mutation `setMembershipAutoPay` is called
- Add mutation `resetMembership` to reset a membership: deactivates membership, restores voucher code, removes periods, sets pledger as user


servers/republik/modules/crowdfundings/graphql/resolvers/_mutations/resetMembership.js
```diff
@@ -0,0 +1,73 @@
...
+        updatedAt: new Date()
+      }
+    )
+
+    await transaction.transactionCommit()
+
+    const { active, id, sequenceNumber, userId } = updatedMembership
+
+    try {
+      if (active) {
```
tpreusse, line 49: why would active ever be true? The `updateAndGetOne` explicitly sets active to `false`. Should the `enforceSubscriptions` be on the original owner (`membership.userId`)? (in cases that one no longer has an active membership and needs updated customer record on MC)

patrickvenetz: This can't ever be true, this is a horribly piece of code. Changing.

PR #482, opened 2020-08-12
fix: don't try to send failed mails more than 5 times per 24h

patte:


caillou:
Looks good to me. Not tested.

PR #484, opened 2020-08-13
Immediately cancel grace membership

caillou:


servers/republik/modules/crowdfundings/lib/generateMemberships.js
```diff
@@ -260,3 +262,11 @@ module.exports = async (pledgeId, pgdb, t, redis) => {
   const cache = createCache({ prefix: `User:${user.id}` }, { redis })
   cache.invalidate()
 }
+
+async function isMembershipInGracePeriod (membership, pgdb) {
+  const hasActivePeriod = await pgdb.public.membershipPeriods.count({
+    membershipId: membership.id,
+    'endDate >': new Date()
+  })
+  return !hasActivePeriod
```
patte, line 271:
```suggestion
  const hasPeriodEndingInTheFuture = await pgdb.public.membershipPeriods.count({
    membershipId: membership.id,
    'endDate >': new Date()
  })
  return !hasPeriodEndingInTheFuture
```

tpreusse:
One input: wouldn't it be better if immediately would always be set to true in the cancel method?
https://github.com/orbiting/backends/blob/54025684ae123153feb4cc36ce21209c36bc284f/servers/republik/modules/crowdfundings/lib/cancelMembership.js#L40-L47

For example when a user cancels instead of updating the credit card. Currently the frontend never sends `immediately`.

patte:
@tpreusse I'm not sure I understand. Immediately cancelling a membership with a period ending in the future leads to loosing the days between `now` and then. Which is not what most customers want. Can you explain what you mean?

caillou:
@tpreusse could we then remove this line: https://github.com/orbiting/backends/blob/master/servers/republik/modules/crowdfundings/graphql/schema.js#L66

tpreusse:
> @tpreusse could we then remove this line: https://github.com/orbiting/backends/blob/master/servers/republik/modules/crowdfundings/graphql/schema.js#L66

No because [republik-admin-frontend](https://github.com/orbiting/republik-admin-frontend) does use it.

caillou:
@patte I just deployed the current state and did a manual test. Could you review one more time, just in case?

servers/republik/modules/crowdfundings/graphql/resolvers/Membership.js
```diff
@@ -29,19 +29,25 @@ const membershipResolver = {
...
 
-    const latest = await pgdb.public.membershipPeriods.findFirst({
-      membershipId: membership.id
-    }, { fields: '"endDate"', orderBy: ['endDate desc'] })
-    return !!(
-      membership.active &&
-      membership.latestPaymentFailedAt &&
-      membership.latestPaymentFailedAt > latest.endDate
-    )
+    const latestPeriod = periods[periods.length - 1]
```
patte, line 46:
```suggestion
    const latestPeriod = periods[0]
```
periods is sorted `DESC`.

servers/republik/modules/crowdfundings/graphql/resolvers/Membership.js
```diff
@@ -29,19 +29,25 @@ const membershipResolver = {
...
+    const isLatestPeriodEnded = new Date(latestPeriod.endDate) < new Date()
+    return isLatestPeriodEnded
```
patte, line 48:
```suggestion
    return new Date(latestPeriod.endDate) < new Date()
```

servers/republik/modules/crowdfundings/lib/cancelMembership.js
```diff
@@ -2,9 +2,15 @@ const debug = require('debug')('crowdfundings:cancelMembership')
...
 const cancelSubscription = require('./payments/stripe/cancelSubscription')
 
+const { overdue } = require('../graphql/resolvers/Membership')
+
 module.exports = async (membership, details, options, t, pgdb) => {
   const { reason, type, suppressConfirmation, suppressWinback, cancelledViaSupport } = details
-  const { immediately } = options
+  let { immediately } = options
+
+  if (overdue(membership, null, { pgdb })) {
```
patte, line 11:
```suggestion
  if (!immediately && await overdue(membership, null, { pgdb })) {
```
:point_up:

PR #486, opened 2020-08-14
fix(mail): rm brackets[] from plain text anchors, add space after urls

tpreusse:
```diff
Haben Sie einen Zugriff auf republik.ch angefordert?
-Dann klicken Sie bitte auf den folgenden Link: 
+Dann klicken Sie bitte auf den folgenden Link. 

-Zugriff bestätigen [{{LOGIN_LINK}}]
+Zugriff bestätigen: 
+{{LOGIN_LINK}} 

-Falls Sie diese Zugriffsanfrage nicht ausgelöst haben, können Sie die
-Zugriffsaufforderung ablehnen:
-
-[{{LOGIN_LINK}}&noAutoAuthorize=true]
+Falls Sie diese Zugriffsanfrage nicht ausgelöst haben, 
+können Sie die Zugriffsaufforderung ablehnen: 
+{{LOGIN_LINK}}&noAutoAuthorize=true 

```

### DIFF
--- a/packages/mail/script/generateTextTemplates.js
+++ b/packages/mail/script/generateTextTemplates.js
@@ -1,6 +1,7 @@
 #!/usr/bin/env node
 const { promises: fs } = require('fs')
 const htmlToText = require('html-to-text')
+const defaultFormat = require('html-to-text/lib/formatter')
 const path = require('path')
 const Promise = require('bluebird')
 const recursive = require('recursive-readdir')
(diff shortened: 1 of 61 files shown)


packages/mail/script/generateTextTemplates.js
```diff
@@ -30,13 +31,59 @@ const run = async () => {
...
+                }
+
+                const storedCharCount = options.lineCharCount
+                const text = getText()
+                const href = getHref()
+                function normalizeUrl (url = '') {
+                  return url
+                    .replace(/\s+/g, '')
+                    .replace(/^(https?:\/\/)?(www\.)?/, '')
+                    .replace('{{frontend_base_url}}', 'republik.ch')
```
patte, line 62: why is this needed?

packages/mail/script/generateTextTemplates.js
```diff
@@ -30,13 +31,59 @@ const run = async () => {
       await fs.writeFile(
         file.replace(/\.html$/gi, '.txt'),
         htmlToText
-          .fromString(html, { ignoreImage: true })
-          .replace(/^ /gm, '')
-          .replace(/kontakt@republik\.ch \[kontakt@republik\.ch\]/gm, 'kontakt@republik.ch'),
```
patte, line 35: this is not needed anymore?

packages/mail/script/generateTextTemplates.js
tpreusse: to avoid `www.republik.ch {{frontend_base_url}} ` which would result in `www.republik.ch https://www.republik.ch/ `

Now it will just generate `{{frontend_base_url}}` which results in `https://www.republik.ch/ ` (hopefully clickable)

packages/mail/script/generateTextTemplates.js
tpreusse: it's handled by `toHideSameLink` (aka `hideLinkHrefIfSameAsText` option which is baked into our custom format anchor)

patte: yes, yes, after understanding the first question, I now also see this.

PR #488, opened 2020-08-19
fix: quotes in notification titles

patte:
Fixes double quotes in notification titles, e.g.
CED3027B-7887-4592-A087-B49080A3473C


caillou (approved):
LGTM

PR #493, opened 2020-08-21
Initial Typescript Setup

caillou:
# Initial Typescript Setup

## Abstract

The aim of this PR is to have a working and deployable TypeScript setup. 

In addition, we would like to experiment with `JS` to `TS` migration. For that purpose, the following file and all it's dependencies should be migrated to TypeScript. The `any` type should be avoided, with the exception if highly generic code.

## Tasks

- [x] Migrate the following files to TS
  - [x] packages/auth/loaders/User.ts
  - [x] packages/auth/lib/transformUser.ts
  - [x] packages/auth/lib/index.ts
  - [x] packages/auth/index.ts
  - [x] packages/dataloader/index.ts
- [x] Set up `TS`:
  - [x] `tsconfig.json`
  - [x] add dependencies 
  - [x] build task with watcher
  - [x] nodemon for dev
- [x] Deployment


tpreusse (approved):
Looks good 🎉 

One thing: we should add [eslint & prettier](https://www.robertcooper.me/using-eslint-and-prettier-in-a-typescript-project) for tsc files. Currently the style is inconsistent and a few semicolons snuck in. But IMO something for a separate PR.
Would be

patte:
`yarn run tsc` gives me the following error:
```
yarn run v1.22.4
$ /home/patte/republik/src/backends/node_modules/.bin/tsc
packages/auth/loaders/User.ts:48:15 - error TS1005: ',' expected.

48     credential: createDataLoader(ids =>
                 ~

packages/auth/loaders/User.ts:51:3 - error TS1005: ',' expected.

51   }
     ~


Found 2 errors.
```
Also, could you please add a section to the `README.md` explaining that the project uses typescript, how to run in development, about not to check in generated files and how it's run in prod.

packages/auth/lib/index.ts
```diff
@@ -1,4 +1,6 @@
-module.exports = {
+import transfromUser = require('./transformUser')
+
+export default {
```
tpreusse, line 3: I get the following error on staging:
Screenshot 2020-08-29 at 16 56 54

Caused by the `revokeConsent` mutation:
https://github.com/orbiting/backends/blob/458cf43e19724854f56b398485475fe8da52786d/packages/auth/graphql/resolvers/_mutations/revokeConsent.js#L1

seems like requiring from this file after changing export style does not work as expected.

tpreusse (changes requested):

tsconfig.json
```diff
@@ -0,0 +1,70 @@
...
+    // "composite": true, /* Enable project compilation */
+    // "tsBuildInfoFile": "./", /* Specify file to store incremental compilation information */
+    // "removeComments": true, /* Do not emit comments to output. */
+    // "noEmit": true, /* Do not emit outputs. */
+    // "importHelpers": true, /* Import emit helpers from 'tslib'. */
+    // "downlevelIteration": true, /* Provide full support for iterables in 'for-of', spread, and destructuring when targeting 'ES5' or 'ES3'. */
+    // "isolatedModules": true, /* Transpile each file as a separate module (similar to 'ts.transpileModule'). */
+
+    /* Strict Type-Checking Options */
+    // "strictNullChecks": true, /* Enable strict null checks. */
```
philippstucki, line 38: I highly recommend enabling strict null checks.

caillou:
@tpreusse ready for review once again.

tpreusse (approved):

PR #494, opened 2020-08-24
refactor: extract modules/crowdfundings into a package

patte:


packages/utils/package.json
```diff
@@ -24,6 +24,7 @@
     "bluebird": "^3.7.2",
     "debug": "^4.1.1",
     "get-urls": "^9.2.0",
-    "moment": "^2.26.0"
+    "moment": "^2.26.0",
+    "unified": "^9.0.0"
```
caillou, line 28: Now this could also be removed from `servers/graphql/package.json`, right?

caillou:
Did you consider moving the `crowdfunding` migraion files too?

I don't think it is needed, really, would just make things a wee bit more consisten …

packages/republik/graphql/index.js
```diff
@@ -4,6 +4,5 @@ const { graphql: scalars } = require('@orbiting/backend-modules-scalars')
 const local = addTypes(loadModule(__dirname), [scalars])
 
 const { graphql: auth } = require('@orbiting/backend-modules-auth')
```
caillou, line 6: One thing I noticed, is that we still require, but never use `const { graphql: auth } = require('@orbiting/backend-modules-auth')` in the following file: `servers/graphql/server.js`. I think that was forgotten when we merged the two servers.

caillou (approved):
Looks good to me. I added three comments, which are not crucial in my opinion!

Thanks for cleaning up all of this!

PR #495, opened 2020-08-26
fix: merge stripe customers and fix reclaimPledge, mergeUsers

patte:
The important lib: `packages/republik-crowdfundings/lib/payments/stripe/mergeCustomers.js`

### DIFF
--- a/packages/republik-crowdfundings/lib/payments/stripe/mergeCustomers.js
+++ b/packages/republik-crowdfundings/lib/payments/stripe/mergeCustomers.js
@@ -75,8 +75,11 @@ module.exports = async ({
   }
 
   // check sources
-  const getSources = (customers) =>
-    customers.reduce( (acc, cur) => acc.concat(...cur.sources.data), [])
+  const getSources = (customers) => customers.reduce( (acc, cur) => acc.concat(
+    ...cur.default_source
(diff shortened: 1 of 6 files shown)


packages/republik-crowdfundings/lib/payments/stripe/mergeCustomers.js
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  if (hasSubscriptions(sourceCustomers)) {
+    debug('only source has subscriptions, move source to target')
+    return moveStripeCustomers(sourceUserId, targetUserId)
+  }
+
+  // check sources
+  const getSources = (customers) =>
+    customers.reduce( (acc, cur) => acc.concat(...cur.sources.data), [])
```
tpreusse, line 79: I'd limit this to the default card because (rename to `getDefaultSources`) because we won't use other non default cards (e.g. from a book purchase) for autopay or even next purchases.

tpreusse (commented):
Generally looks good. I think we should limit source max exp eval to default cards.

packages/republik-crowdfundings/lib/payments/stripe/mergeCustomers.js
patte: thanks for the input. applied.

PR #499, opened 2020-08-28
Anonymous comment differentiator

ruggedly:
Done: 
* If user sets comments to anonymous an anonymous diff. is added to db entry
* Differentiator will be displayed at comments

TODO:
* edge cases

### DIFF
--- a/packages/discussions/graphql/resolvers/Comment.js
+++ b/packages/discussions/graphql/resolvers/Comment.js
@@ -242,12 +242,13 @@ module.exports = {
     } else { // FORBIDDEN or ALLOWED
       if (commenterPreferences && commenterPreferences.anonymous != null) {
         anonymous = commenterPreferences.anonymous
-        if (anonymous && commenterPreferences.anonymousDifferentiator !== null) {
-          anonymousComment.name += ` ${commenterPreferences.anonymousDifferentiator}`
-        }
       } else {
         anonymous = false
       }
+    }
+
+    if (anonymous && commenterPreferences.anonymousDifferentiator !== null) {
+      anonymousComment.name += ` ${commenterPreferences.anonymousDifferentiator}`
     }
 
     const profilePicture = getPortrait(commenter, (args && args.portrait), context)
--- a/packages/discussions/graphql/resolvers/Comment.js
+++ b/packages/discussions/graphql/resolvers/Comment.js
@@ -242,6 +242,9 @@ module.exports = {
     } else { // FORBIDDEN or ALLOWED
       if (commenterPreferences && commenterPreferences.anonymous != null) {
         anonymous = commenterPreferences.anonymous
+        if (anonymous && commenterPreferences.anonymousDifferentiator !== null) {
+          anonymousComment.name += ` ${commenterPreferences.anonymousDifferentiator}`
+        }
       } else {
         anonymous = false
       }
--- a/packages/discussions/lib/discussionPreferences.js
+++ b/packages/discussions/lib/discussionPreferences.js
@@ -109,7 +109,7 @@ const ensureAnonymousDifferentiator = async ({
     return
   }
 
-  const userHasComments = !!await transaction.public.comments.findOne({
+  const userHasComments = !!await transaction.public.comments.findFirst({
     userId, discussionId
   })
 
--- a/packages/discussions/lib/discussionPreferences.js
+++ b/packages/discussions/lib/discussionPreferences.js
@@ -78,17 +78,21 @@ const setDiscussionPreferences = async ({
     skipUndefined: true
   }
 
+  let mutation
+
   if (existingDP) {
-    await transaction.public.discussionPreferences.updateOne(findQuery, updateQuery, options)
+    mutation = transaction.public.discussionPreferences.updateOne(findQuery, updateQuery, options)
   } else {
-    await Promise.all([
-      transaction.public.discussionPreferences.insert(updateQuery, options),
-      loaders.Discussion.Commenter.discussionPreferences.clear({
-        userId: userId,
-        discussionId: discussion.id
-      })
-    ])
+    mutation = transaction.public.discussionPreferences.insert(updateQuery, options)
   }
+
+  await Promise.all([
+    mutation,
+    loaders.Discussion.Commenter.discussionPreferences.clear({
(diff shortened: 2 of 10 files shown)


packages/discussions/lib/setDiscussionPreferences.js
```diff
@@ -61,10 +61,22 @@ module.exports = async ({
...
+  if (anonymity) {
+    if (!existingDP || existingDP.anonymousDifferentiator === null) {
+      const lastUsed = await transaction.public.discussionPreferences.findOne({
+        discussionId: discussion.id,
+        'anonymousDifferentiator !=': null
+      }, { orderBy: { anonymousDifferentiator: 'DESC' } })
+      anonymousDiff = lastUsed ? lastUsed.anonymousDifferentiator + 1 : 1
+    }
+  }
+
```
tpreusse, line 74: I would recommend to move this logic to `submitComment`. By doing so we could ensure that the `anonymousDifferentiator` is only assigned once some one actually participates. Additionally we could also handle anonymity enforced debates by creating a `discussionPreference` with `anonymous` and  `anonymousDifferentiator` on submit.

packages/discussions/graphql/resolvers/Comment.js
```diff
@@ -242,6 +242,9 @@ module.exports = {
     } else { // FORBIDDEN or ALLOWED
       if (commenterPreferences && commenterPreferences.anonymous != null) {
         anonymous = commenterPreferences.anonymous
+        if (anonymous && commenterPreferences.anonymousDifferentiator !== null) {
+          anonymousComment.name += ` ${commenterPreferences.anonymousDifferentiator}`
+        }
```
tpreusse, line 247: If `anonymousDifferentiator` is set in `submitComment` we could move this after the whole if else block and also set it for `ENFORCED` discussions.

packages/discussions/lib/setDiscussionPreferences.js
caillou: I see where you are coming from. Unfortunatly, this could lead to some edge-cases that will need to be faught with more code.

Say `Ada Lovelace` has already commented in a thread. If she now changes her anonymity settings *a posteriori* without submitting a new comment, the `anonymousDifferentiator` will not be set.

I'll try to figure out a clean way.

tpreusse: Ah true that. Then I would recommend leaving this here but only setting `anonymousDifferentiator` when:

- it is called from `submitComment` with a new special arg `setAnonymousDifferentiator` where we'd now always call it (potential optimise to avoid unnecessary mutations when a `anonymousDifferentiator` is not needed)
- or when not called from `submitComment` / with `setAnonymousDifferentiator` but already has a comment (`comments.find({userId, discussionId: discussion.id })`)

caillou:
@tpreusse ready for review…

packages/discussions/lib/discussionPreferences.js
```diff
@@ -80,6 +81,92 @@ module.exports = async ({
   if (existingDP) {
     await transaction.public.discussionPreferences.updateOne(findQuery, updateQuery, options)
   } else {
-    await transaction.public.discussionPreferences.insert(updateQuery, options)
+    await Promise.all([
+      transaction.public.discussionPreferences.insert(updateQuery, options),
+      loaders.Discussion.Commenter.discussionPreferences.clear({
+        userId: userId,
+        discussionId: discussion.id
+      })
```
tpreusse, line 89: What the reason that this only needs to happen on `insert` and not after `updateOne`? IMO could be moved after the `if (existingDP)` block.

packages/discussions/lib/discussionPreferences.js
```diff
@@ -80,6 +81,92 @@ module.exports = async ({
...
+
+  const preferences = await transaction.public.discussionPreferences.findOne(findQuery)
+
+  if (preferences && preferences.anonymousDifferentiator) {
+    return
+  }
+
+  const userHasComments = !!await transaction.public.comments.findFirst({
+    userId, discussionId
+  })
```
tpreusse, line 123:
```suggestion
  const userHasComments = await transaction.public.comments.count({
    userId, discussionId
  }) > 0
```

tpreusse (approved):
Nice. I have one improvement note and a question. Otherwise good to go from my side. Also approved by CR to go out without communication.

packages/discussions/lib/discussionPreferences.js
caillou: Argh, this is a mistake.

I only wanted to clear the cache on `update` (and I actually went for `insert`), but of course, an empty response would be cached in the loader too.

packages/discussions/lib/discussionPreferences.js
caillou: Is this a matter of style?

I feel, that as someone who reads the code, it is more likely to miss the `> 0` at the end of a line than the `!!` at the beginning of a line.

Anyway, not agreing on how to write the type cast, tells me that I should go for this:

```js
const amountOfComments = await transaction.public.comments.count({
   userId, discussionId
})

if (amountOfComments === 0) {
  return
}
```

tpreusse: I'm also fine with
```suggestion
  const userHasComments = !!await transaction.public.comments.count({
    userId, discussionId
  })
```

My main suggestion is using `count` instead of `findFirst` since we don't need the data.

caillou:
@tpreusse ready for review once again.

packages/discussions/lib/discussionPreferences.js
```diff
@@ -77,9 +78,99 @@ module.exports = async ({
...
-    await transaction.public.discussionPreferences.insert(updateQuery, options)
+    mutation = transaction.public.discussionPreferences.insert(updateQuery, options)
+  }
+
+  await Promise.all([
+    mutation,
+    loaders.Discussion.Commenter.discussionPreferences.clear({
+      userId: userId,
+      discussionId: discussion.id
+    })
```
tpreusse, line 94:
```suggestion
    loaders.Discussion.Commenter.discussionPreferences.clear(findQuery)
```

tpreusse (approved):
LGTM

PR #503, opened 2020-09-07
Fix missing await

caillou:
Fix missing await and temporarily comment out critical code, as this has not been tested.

An issue has been created, in order not to forget to un-comment the critical code: https://github.com/orbiting/backends/issues/504


tpreusse (approved):
let's try it 🙈

PR #509, opened 2020-09-11
Collection Items

tpreusse:
Allowing to query `collectionItems` from multiple collections at the same time.

Example query:
```
{
  me {
    collectionItems(names: ["bookmarks", "progress"], first: 10) {
      totalCount
      nodes {
        id
        collection {
          name
        }
        document {
          meta {
            title
          }
        }
      }
    }
  }
}
```

### DIFF
--- a/packages/collections/graphql/resolvers/User.js
+++ b/packages/collections/graphql/resolvers/User.js
@@ -5,32 +5,28 @@ const { paginate } = require('@orbiting/backend-modules-utils')
 const accessRoles = ['member']
 const adminRoles = ['admin', 'supporter']
 
+const canAccess = (user, context) => {
+  const { user: me } = context
+  return (
+    (Roles.userIsMe(user, me) && Roles.userIsInRoles(user, accessRoles)) ||
+    Roles.userIsInRoles(me, adminRoles)
+  )
+}
+
 module.exports = {
   collections (user, args, context) {
-    const { user: me } = context
-    if (
-      (Roles.userIsMe(user, me) && Roles.userIsInRoles(user, accessRoles)) ||
-      Roles.userIsInRoles(me, adminRoles)
-    ) {
+    if (canAccess(user, context)) {
       return Collection.findForUser(user.id, context)
     }
     return []
   },
   collection (user, { name }, context) {
-    const { user: me } = context
-    if (
-      (Roles.userIsMe(user, me) && Roles.userIsInRoles(user, accessRoles)) ||
-      Roles.userIsInRoles(me, adminRoles)
-    ) {
+    if (canAccess(user, context)) {
       return Collection.byNameForUser(name, user.id, context)
     }
   },
   async collectionItems (user, args, context) {
-    const { user: me } = context
-    if (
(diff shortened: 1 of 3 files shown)


packages/collections/graphql/resolvers/User.js
```diff
@@ -23,5 +24,22 @@ module.exports = {
     ) {
       return Collection.byNameForUser(name, user.id, context)
     }
+  },
+  async collectionItems (user, args, context) {
+    const { user: me } = context
+    if (
+      (Roles.userIsMe(user, me) && Roles.userIsInRoles(user, accessRoles)) ||
+      Roles.userIsInRoles(me, adminRoles)
+    ) {
```
caillou, line 33: Could you DRY this code?

Something like:

```js
function mayAccess ({ user, me }) {
  return (Roles.userIsMe(user, me) && Roles.userIsInRoles(user, accessRoles)) ||
    Roles.userIsInRoles(me, adminRoles)
}
```

packages/collections/graphql/resolvers/User.js
```diff
@@ -23,5 +24,22 @@ module.exports = {
...
+      const collections = await Promise.all(args.names.map(name => Collection.byNameForUser(name, user.id, context)))
```
caillou, line 34: We could reduce the amount of call to the DB as follows, right?
```suggestion
       const collections = await Collection.findForUser(user.id, context)
      .then(collections => collections.filter(
        collection => args.names.includes(collection.name)
      ))
```

This code was not tested ;)

caillou (changes requested):

packages/collections/graphql/resolvers/User.js
tpreusse: Good point. I've extracted it to an helper now: https://github.com/orbiting/backends/pull/509/commits/de0a328e3de44204cd104edafe2d2204b7818365

packages/collections/graphql/resolvers/User.js
tpreusse: `Collection.byNameForUser` uses a data loader under the hood which gets those collections in one sql query. So the current implementation is more efficient because it only loads the ones that are requested and also only does one db query.

caillou:
LGTM now!

packages/collections/graphql/resolvers/User.js
```diff
@@ -1,27 +1,42 @@
...
+      const items = await Collection.findDocumentItems(
+        {
+          collectionId: collections.map(collection => collection.id),
+          userId: user.id
+        },
+        context
+      )
+      return paginate(args, items)
+    }
+    return paginate(args, [])
```
caillou, line 40: One detail: I'd like to push towards early return statements, as these are much more readable in my opinion. The `return paginate(args, [])` is much closer to the cause, being `!canAccess()`:

```suggestion
 if (!canAccess(user, context)) paginate(args, [])

const collections = await Promise.all(args.names.map(name => Collection.byNameForUser(name, user.id, context)))
const items = await Collection.findDocumentItems(
  {
    collectionId: collections.map(collection => collection.id),
    userId: user.id
  },
  context
)
return paginate(args, items)
```

PR #511, opened 2020-09-14
User prettier for JS files

caillou:
This PR adds prettier to JS files and runns it for all files.


tpreusse (approved):
LGTM. Assuming everything was done automatically and now manual fixes were applied.

caillou:
> LGTM. Assuming everything was done automatically and no manual fixes were applied.

I manually fixed one line in an HTML file:

https://github.com/orbiting/backends/blob/master/packages/mail/templates/payment_successful_abo_give.html#L38

```diff
- …ch bei uns eingegangen.<strong></p>
+ …ch bei uns eingegangen.</strong></p>
```

Would you agree that this change is safe?

PR #514, opened 2020-09-15
feat(notifications): filter for EventObjectType

tpreusse:
Allow to query for notifications for a specific type.

```gql
{
  notifications(first: 2, filter: Document) {
    totalCount
    nodes {
      object {
        ... on Document {
          id
          meta {
            title
          }
        }
      }
    }
  }
}
```

### DIFF
--- a/packages/subscriptions/graphql/resolvers/_queries/notifications.js
+++ b/packages/subscriptions/graphql/resolvers/_queries/notifications.js
@@ -9,9 +9,9 @@ module.exports = async (_, args, context) => {
 
   const nodes = await pgdb.public.notifications.find(
     {
-      userId: me.id,
       ...(args.onlyUnread ? { readAt: null } : {}),
       ...(args.filter ? { eventObjectType: args.filter } : {}),
+      userId: me.id,
     },
     {
       orderBy: { createdAt: 'DESC' },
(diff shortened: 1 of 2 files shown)


tpreusse:
@caillou hatten für «Meine Republik» noch was vergessen. Gerne reviewen.

packages/subscriptions/graphql/resolvers/_queries/notifications.js
```diff
@@ -11,6 +11,7 @@ module.exports = async (_, args, context) => {
     {
       userId: me.id,
       ...(args.onlyUnread ? { readAt: null } : {}),
+      ...(args.filter ? { eventObjectType: args.filter } : {}),
     },
```
caillou, line 15:
```suggestion
    {
      ...(args.onlyUnread ? { readAt: null } : {}),
      ...(args.filter ? { eventObjectType: args.filter } : {}),
      userId: me.id,
    },
```

This code makes me a wee bit nervous. I had to go and read the definition of the `EventObjectType` to makes sure it does not conatina and overwirte the `userId`.

caillou: And I see how what I wrote makes no senese ;)

Still getting used to using the spread operator to add a conditional property.

PR #518, opened 2020-09-22
Three Small Things

tpreusse:
## progress filter

For continue reading we want to do this:
```gql
{
  me {
    id
    name
    collectionItems(names: ["bookmarks", "progress"], progress: UNFINISHED) {
      totalCount
      nodes {
        document {
          meta {
            title
          }
          userProgress {
            percentage
          }
        }
      }
    }
  }
}

```

And for a new bookmark page we also want to displayed finished items:
```gql
{
  me {
    id
    name
    collectionItems(names: ["progress"], progress: FINISHED) {
      totalCount
      nodes {
        document {
          meta {
            title
          }
          userProgress {
            percentage
          }
        }
      }
    }
  }
}
```

## slack greeting

The bot unfortunately sends messages as attachments. Let's support that. And also ignore threaded messages going forward, was never supported on purpose.

## username of privates profiles

Currently if an author has a private profile we don't resolve the username. E.g. [here](https://www.republik.ch/2020/09/21/die-waechter-aus-dem-weissen-haus). Let's change that.

### DIFF
--- a/packages/collections/graphql/resolvers/User.js
+++ b/packages/collections/graphql/resolvers/User.js
@@ -27,17 +27,10 @@ module.exports = {
   },
   async collectionItems(user, args, context) {
     if (canAccess(user, context)) {
-      const collections = await Promise.all(
-        args.names.map((name) =>
-          Collection.byNameForUser(name, user.id, context),
-        ),
-      )
-
-      const items = await Collection.findDocumentItems(
+      const items = await Collection.findDocumentItemsByCollectionNames(
         {
-          collectionId: collections.map((collection) => collection.id),
+          ...args,
           userId: user.id,
-          ...Collection.getProgressConditions(args.progress),
         },
         context,
       )
--- a/packages/collections/graphql/resolvers/User.js
+++ b/packages/collections/graphql/resolvers/User.js
@@ -13,22 +13,6 @@ const canAccess = (user, context) => {
   )
 }
 
-const progressFinishedThreshold = 0.85
-const progressConditions = {
-  FINISHED: {
-    or: [
-      { "data->>'percentage' >=": progressFinishedThreshold },
-      { "data->>'percentage'": null },
-    ],
-  },
-  UNFINISHED: {
-    or: [
-      { "data->>'percentage' <=": progressFinishedThreshold },
-      { "data->>'percentage'": null },
-    ],
-  },
-}
-
 module.exports = {
   collections(user, args, context) {
     if (canAccess(user, context)) {
@@ -48,11 +32,12 @@ module.exports = {
           Collection.byNameForUser(name, user.id, context),
         ),
       )
+
       const items = await Collection.findDocumentItems(
         {
           collectionId: collections.map((collection) => collection.id),
           userId: user.id,
-          ...progressConditions[args.progress],
+          ...Collection.getProgressConditions(args.progress),
         },
         context,
       )
--- a/packages/collections/graphql/resolvers/User.js
+++ b/packages/collections/graphql/resolvers/User.js
@@ -13,6 +13,22 @@ const canAccess = (user, context) => {
   )
 }
 
+const progressFinishedThreshold = 0.85
+const progressConditions = {
+  FINISHED: {
+    or: [
+      { "data->>'percentage' >=": progressFinishedThreshold },
+      { "data->>'percentage'": null },
+    ],
+  },
+  UNFINISHED: {
+    or: [
+      { "data->>'percentage' <=": progressFinishedThreshold },
+      { "data->>'percentage'": null },
+    ],
+  },
+}
+
 module.exports = {
   collections(user, args, context) {
     if (canAccess(user, context)) {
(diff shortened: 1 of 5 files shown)


packages/collections/graphql/schema-types.js
```diff
@@ -45,11 +45,17 @@ extend type Document {
   userProgress: DocumentProgress
 }
 
+enum ProgressState {
+  FINISHED
+  UNFINISHED
+}
```
caillou, line 51: My intuition tells me, one state is missing. WDYT?

```suggestion
enum ProgressState {
  NOT_STARTED
  IN_PROGRESS
  COMPLETED
}
```

packages/collections/graphql/resolvers/User.js
```diff
@@ -13,6 +13,22 @@ const canAccess = (user, context) => {
   )
 }
 
+const progressFinishedThreshold = 0.85
+const progressConditions = {
+  FINISHED: {
+    or: [
+      { "data->>'percentage' >=": progressFinishedThreshold },
+      { "data->>'percentage'": null },
```
caillou, line 21: What is the motivation of selecting records `WHERE "percentage" IS NULL`?

packages/collections/graphql/resolvers/User.js
```diff
@@ -13,6 +13,22 @@ const canAccess = (user, context) => {
   )
 }
 
+const progressFinishedThreshold = 0.85
```
caillou, line 16: Where does this magic number come from?

We will probably have to make this configurable over the API at some point, no?

packages/collections/graphql/schema-types.js
tpreusse: `NOT_STARTED` state or deleted positions (`removeDocumentProgress`) are represented by no record being present. We only start recording the position once someone scrolled over the first paragraph of the main text.

caillou: API !== implementation

packages/collections/graphql/resolvers/User.js
tpreusse: The null selector ensure compatibility with non-progress items. E.g. selecting `bookmark` (without a `data->>'percentage'`) and `progress` items that are unfinished.

packages/collections/graphql/schema-types.js
caillou: Ok. I get it now.

packages/collections/graphql/resolvers/User.js
ovbm: I don't know if will ever use this from the API - but it kind of should be consistent with the frontend, where we only show "finished" or "Gelesen" when progress is 1 (100%). So that's a bit inconsistent in my mind.

ovbm:
tested this locally with the new query filter in publikator and it works. Only open question is the threshold.

tpreusse:
> tested this locally with the new query filter in publikator and it works. Only open question is the threshold.

We just discussed this in person and came up with a one week threshhold:
https://github.com/orbiting/backends/pull/518/files#diff-282558113acd49fb56087b69ac654c77R229-R247

Basically ensuring that old stuff with over `0.75` does not popup in the continue reading lists.

ovbm:
My issue with this solution is that someone would look for an article they nearly finished longer than a week ago and they wouldn't find it in "weiterlesen" -> they would filter the list to "Gelesene" and they would find the article there, with 75% done. Which doens't make any sense, as those articles should all show a check-mark for 100% completion.

ovbm:
I don't think we should have any threshold in the backend, as this would create weird inconsistencies with the frontend. It's very easy from a user perspective to mark things as "read", so why make it weird for them to find and manage their bookmarks and user progress? You'd have to explain the threshold and it could be quite confusing.

packages/collections/lib/Collection.js
```diff
@@ -219,7 +220,37 @@ const clearItems = async (userId, collectionName, { pgdb, loaders }) => {
...
+    UNFINISHED: {
+      or: [
+        { "data->>'percentage' <=": PROGRESS_FINISHED_THRESHOLD },
+        { "data->>'percentage' <": 1, 'updatedAt >': aWeekAgo },
+        { "data->>'percentage'": null },
+      ],
+    },
+  }
+  return conditions[progressState]
+}
```
caillou, line 249: Adds documentation and uses a switch to simplify the code.

```suggestion
const getProgressConditions = (progressState) => {
  if (!progressState) {
    return
  }

  // We consider a recently read artcile only as finished, if it was fully
  // read. An article that has not been looked at for 7 days is considered 
  // read, if more than the threshold (75%) was read.

  const aWeekAgo = moment().subtract(7, 'days')

  switch (progressState) {
    case 'FINISHED':
      return {
        or: [
          {
            "data->>'percentage' >=": PROGRESS_FINISHED_THRESHOLD,
            'updatedAt <=': aWeekAgo,
          },
          { "data->>'percentage'": 1 },
          { "data->>'percentage'": null },
        ],
      }

    case 'UNFINISHED':
      return {
        or: [
          { "data->>'percentage' <=": PROGRESS_FINISHED_THRESHOLD },
          { "data->>'percentage' <": 1, 'updatedAt >': aWeekAgo },
          { "data->>'percentage'": null },
        ],
      }
  }
}
```

PR #519, opened 2020-09-29
Validate presence of title and slug in template repo

annatraussnig:
Translations missing - can you grant me access to the backend translations file @tpreusse ?

### DIFF
--- a/packages/publikator/graphql/resolvers/_mutations/commit.js
+++ b/packages/publikator/graphql/resolvers/_mutations/commit.js
@@ -112,11 +112,11 @@ module.exports = async (_, args, context) => {
   let repo = await getRepo(repoId).catch((response) => null)
 
   if (isTemplate && !mdast.meta.title) {
-    throw new Error('Title required for templates')
+    throw new Error(t('api/commit/templateTitle/required'))
   }
 
   if (isTemplate && !mdast.meta.slug) {
-    throw new Error('Repo-slug required for templates')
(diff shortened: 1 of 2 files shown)


packages/publikator/graphql/resolvers/_mutations/commit.js
```diff
@@ -111,6 +111,14 @@ module.exports = async (_, args, context) => {
   // get / create repo
   let repo = await getRepo(repoId).catch((response) => null)
 
+  if (isTemplate && !mdast.meta.title) {
```
caillou, line 114: As far as I see, this only concerns new documents. Wouldn't it make sense to move these checks into the `else` clause, just before `repo = await githubRest.repos.createInOrg()`.

caillou: Maybe we could start to refactor this fuction a wee bit?

Something along the lines of

```js
const repo = createOrGetRepo(...)
```

annatraussnig: No it's also on existing repos. That's why I put this outside of the if-else. 
Basically if a repo is a so-called template repo, it must have a title and a slug in its metadata. Otherwise a user could create a repo (with title/slug), then remove either of those params in a subsequent commit without it raising an error.

tpreusse (approved):
Gerne noch kurz ins «backends translations» übertragen

PR #523, opened 2020-10-14
Send payment reminders

caillou:
- [ ] Write `packages/republik-crowdfundings/lib/payments/sendPaymentReminders()`:
  - [ ] If there are umatched payements:
    - [ ] If payments have not been matched for more than `n` days, warn acoountants.
    - [x] `return`
  - [x] `env` variable for `dryMode`.
  - [x] Payments overdue more than `30` days: Send 1st reminder.
  - [x] Payments overdue more than `60` days: Send 2nd reminder.
  - [x] Payments overdue more than `75` days: Inform accountants about pledges to cancel.
  - [x] Inform acountants about the result.
- [x] Call `sendPaymentReminders()` after `matchPayments()`.
- [x] Call `sendPaymentReminders()` after `manuallyMatchPostfinancePayment()`.
- [x]  Update `sendPaymentReminders` mutation to take just one option: `dryMode`, overrides `env`.


tpreusse:
> There are multiple constants for the 30 days used for the dueDate. Shouldn't all these use the same const?

Should be two: `PAYMENT_DEADLINE_DAYS` used for due date. And a `daysTillFirstReminder` (also make it all caps) which actually should be `PAYMENT_DEADLINE_DAYS + 3` to ensure payments could arrive over the weekend.

> How sould the report look like?

Report looks good

> How should the report look, if dryMode === true?

Prefix with a message, e.g.:
> ⚠️ DRY MODE, REMINDERS ARE NOT SENT ⚠️

> Should a transaction be expebcted in the context of sendPaymentReminders()? C.f.

As far as I can see not necessary. Transactions are important if we mutate multiple tables which contacted data.

> Should proceed if there are error in the for … in loop of sendReminders(), c.f.

No. Just report all errors while running to slack so we can investigate.

packages/republik-crowdfundings/lib/payments/paymentslip/sendPaymentReminders.ts
```diff
@@ -0,0 +1,312 @@
...
+  isLast: boolean
+}
+async function send({
+  payment,
+  isLast,
+  context,
+}: SendParameters): Promise<void> {
+  const emailSubject = isLast
+    ? 'LETZTE MAHNUNG Republik-Abonnement/Mitgliedschaft/Spende'
+    : 'Zahlungserinnerung Republik-Abonnement/Mitgliedschaft/Spende'
```
tpreusse, line 247: let's query `packages.name` and differentiate base on that

packages/republik-crowdfundings/lib/payments/paymentslip/sendPaymentReminders.ts
```diff
@@ -0,0 +1,317 @@
...
+      payment.packageName
+    }/subject`,
+    `api/email/payment/reminder${isLast ? '/last' : ''}/default/subject`,
+  ])
+
+  await sendMailTemplate(
+    {
+      to: payment.email,
+      fromEmail: process.env.DEFAULT_MAIL_FROM_ADDRESS,
+      subject: emailSubject, // todo: move strings to `t`
```
tpreusse, line 257:
```suggestion
      subject: emailSubject,
```

tpreusse (approved):

PR #527, opened 2020-10-19
Save postfinance files regardless of errors

caillou:


packages/republik-crowdfundings/lib/scheduler/importPayments.ts
```diff
@@ -37,32 +34,45 @@ export async function importPayments(
...
+    await informFailed(
       `importPayments(): postfinance sync failed with the following error: ${
         e.message || e
       } ${e.stack || ''}`,
     )
     return
   }
+}
+
+const _importPayments = (context: Context) => async (transaction: PgDb) => {
```
tpreusse, line 51: maybe `createPaymentImporter` would be a good name (since the function is a factory)

packages/republik-crowdfundings/lib/scheduler/importPayments.ts
```diff
@@ -37,32 +34,45 @@ export async function importPayments(
...
+  const insertPaymentReport = await insertNewPayments(transaction)
+  const matchingReport = await tryMatchingPayments(transaction, context)
+  const unmatchedPaymentsAmount = await getUnmatchedPaymentsAmount(transaction)
 
   await notifyAccountants({
```
tpreusse, line 56: Now the transaction would be rolled back when `notifyAccountants` fails. Should be moved outside the transaction again.

tpreusse (approved):

PR #531, opened 2020-11-05
fix(auth): Filter User.roles

patrickvenetz:
Returns only exposable roles as defined in [lib/Roles](https://github.com/orbiting/backends/compare/pae/fix-filtered-user-roles?expand=1#diff-92181c7cbf69ecc6d962e11749cbfd07b2ba631beb5e5797b5e319fe3d21bd8dR5-R16). This omits undefined roles like "foobar".

### DIFF
--- a/packages/auth/graphql/resolvers/User.js
+++ b/packages/auth/graphql/resolvers/User.js
@@ -32,8 +32,7 @@ module.exports = {
   },
   roles(user, args, { user: me }) {
     if (Roles.userIsMeOrInRoles(user, me, ['admin', 'supporter'])) {
-      const roles = user.roles ?? []
-      return roles.filter((role) => Roles.exposableRoles.includes(role))
+      return user.roles.filter((role) => Roles.exposableRoles.includes(role))
     }
     return []
   },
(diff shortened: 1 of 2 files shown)


packages/auth/graphql/resolvers/User.js
```diff
@@ -32,7 +32,8 @@ module.exports = {
   },
   roles(user, args, { user: me }) {
     if (Roles.userIsMeOrInRoles(user, me, ['admin', 'supporter'])) {
-      return user.roles
+      const roles = user.roles ?? []
+      return roles.filter((role) => Roles.exposableRoles.includes(role))
```
tpreusse, line 36: Isn't that unnecessary because we already do it in [transformUser](https://github.com/orbiting/backends/blob/master/packages/auth/lib/transformUser.ts#L36)?
```suggestion
      return user.roles.filter((role) => Roles.exposableRoles.includes(role))
```

Since the schema mandates `[String!]!` we would know if it often returned null. And this should always be getting transformed users.

tpreusse (approved):

packages/auth/graphql/resolvers/User.js
patrickvenetz: That is unnecessary. As far as I can tell all `User` wander through `transformUser`.

PR #532, opened 2020-11-10
feat(republik-crowdfundings): Declare postal delivery for a package option

patrickvenetz:
Add column `postalDelivery` to `packageOptions` as a boolean. Expose column on GraphQL type `PackageOption` as `postalDelivery`.

Also migrate all options with goodies to have postalDelivery flag set truthy.


packages/republik-crowdfundings/graphql/schema-types.js
```diff
@@ -101,6 +101,9 @@ type PackageOption {
   additionalPeriods: [MembershipPeriod!]
 
   accessGranted: Boolean
+
+  # If true, option reward is delivered
+  postalDelivery: Boolean!
```
tpreusse, line 106: hmm, I'm leaning to a more generic naming, e.g. `requireAddress`, that way we could also use it for memberships that require addresses:
```suggestion
  requireAddress: Boolean!
```

`requireAddress` would also match what we end up sending up (`address` in `PledgePaymentInput`)

patrickvenetz: Seconding that, with this suggestion:

Moving prop into types `Goodie`, `MembershipType`. Requiring address is less of a package option flag but a reward «declaration».

This would make accessing it a level further below `packageOption.reward.requireAddress`.

patrickvenetz: Adopted.

patrickvenetz: Query like so:

```gql
{
  crowdfunding(name:"TEST") {
    packages {
      options {
        reward {
          ...on MembershipType {
            requireAddress
          }
          __typename
          ...on Goodie {
            requireAddress
          }
        }
      }
    }
  }
}
```

tpreusse (approved):

PR #535, opened 2020-11-16
Fix Postfinance import

patrickvenetz:
This Pull Request:
- [prevents Array.reduce error if no payments ought to be inserted](https://github.com/orbiting/backends/commit/3895b54c55dff12a5b2f3669e6351e6c10a0ce13)
- [removes call-to-action on highly overdue payments](https://github.com/orbiting/backends/commit/0ecb1cf5ab1287d25b22e3759b8dbd89a5fb7044)

### DIFF
--- a/packages/republik-crowdfundings/lib/payments/paymentslip/sendPaymentReminders.ts
+++ b/packages/republik-crowdfundings/lib/payments/paymentslip/sendPaymentReminders.ts
@@ -17,8 +17,7 @@ const DRY_RUN = process.env.DRY_RUN_SEND_PAYMENT_REMINDERS === 'true'
 
 const FIRST_REMINDER_DEADLINE_DAYS = PAYMENT_DEADLINE_DAYS + 3 // 3 days to consider weekends
 const SECOND_REMINDER_DEADLINE_DAYS = FIRST_REMINDER_DEADLINE_DAYS + 30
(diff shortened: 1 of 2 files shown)


tpreusse (approved):
LGTM.

I personally would have kept a cancel count. But I guess ultra dashboard is pretty good at counts too :)

### One more thing

Could we suppress send reminder runs unless there has been an import the same day? We can check for that by looking for an entry in `postfinanceImports` with `createdAt` today. Should be done in the runner in the morning and the one at 12:00.

This should reduce cases where the reminder arrives after the payment is issued but still being processed.

patrickvenetz:
@caillou 

> I wonder why this did not generate a TS error in the first place?

Good question, I shall look into that.

patrickvenetz:
@tpreusse 

> One more thing

I understand your intent and shall deliver something feisty in a subsequent pull request.

caillou:
> @caillou
> 
> > I wonder why this did not generate a TS error in the first place?
> 
> Good question, I shall look into that.

You will end up sad: https://github.com/microsoft/TypeScript/issues/34736

patrickvenetz:
> > @caillou
> > > I wonder why this did not generate a TS error in the first place?
> > 
> > 
> > Good question, I shall look into that.
> 
> You will end up sad: [microsoft/TypeScript#34736](https://github.com/microsoft/TypeScript/issues/34736)

Fair enough.

PR #536, opened 2020-11-16
Postfinance reminder prevent

patrickvenetz:
This Pull Requests prevents sending payment reminders if most recent Postfinance payment import is older than 12 hours.


tpreusse (approved):
LGTM

PR #544, opened 2020-12-10
feat(documents): Process node modifiers

patrickvenetz:
This Pull Request is probably my most accomplished achievement.

A reviewer might want to freshen-up on [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) is it is used extensivly.

This Dynamic Component config …

```json
{
  "props": {
    "foo": "bar"
  },
  "src": "url-to-component",
  "modifiers": [
    {
      "name": "mergePropsWhenUserHasSomeRole",
      "props": {
        "additional": "additional-prop-when-user-has-member-role"
      },
      "roles": [
        "member"
      ]
    }
  ]
}
```

… should return when user is in member role:

```json
{
  "props": {
    "foo": "bar",
    "additional": "additional-prop-when-user-has-member-role"
  },
  "src": "url-to-component"
}
```

… should return when user has no member role:

```json
{
  "props": {
    "foo": "bar"
  },
  "src": "url-to-component"
}
```

### DIFF
--- a/packages/documents/graphql/resolvers/Document.js
+++ b/packages/documents/graphql/resolvers/Document.js
@@ -6,6 +6,7 @@ const {
   processRepoImageUrlsInMeta,
   processImageUrlsInContent,
   processEmbedsInContent,
+  processNodeModifiersInContent,
 } = require('../../lib/process')
 const { getMeta } = require('../../lib/meta')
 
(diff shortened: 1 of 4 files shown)


ovbm (approved):
the world will never be the same.

PR #550, opened 2020-12-18
refactor(discussions): Replace ban environment variable w/ suspensions table

patrickvenetz:
This Pull Request ought to replace `DISCUSSION_BANS` environment variable.

It allows to suspend via GraphQL mutations if current user has admin or supporter role:

```gql
mutation {
  suspendUser(id:"some-user-uuid" until:"2021-05-05") {
    suspendedUntil
  }
}
```

One might also want to revert a suspension prematurly with `unsuspendUser`:

```gql
mutation {
  unsuspendUser(id:"some-user-uuid") {
    isSuspended
  }
}
```

A suspended user is unable to post a new comment.

User data is extend with these props, accessible for user itself or user with admin or supporter role:

- `User.isSuspended: Boolean`, indicates whether or not a user is currently suspended
- `User.suspendedUntil: DateTime`, indicates until when a user is suspended
- `User.suspensions: [DiscussionSuspension!]`, returns a list of current suspensions. Use `withInactive: Boolean` argument to return inactive suspensions, too.

### DIFF
--- a/packages/discussions/graphql/resolvers/_mutations/suspendUser.js
+++ b/packages/discussions/graphql/resolvers/_mutations/suspendUser.js
@@ -1,0 +1,25 @@
+const { Roles } = require('@orbiting/backend-modules-auth')
+
+module.exports = async (_, args, context) => {
+  const { pgdb, loaders, user: me } = context
+
+  Roles.ensureUserIsInRoles(me, ['admin', 'supporter'])
+
+  const tx = await pgdb.transactionBegin()
+  try {
+    await pgdb.public.discussionSuspensions.insert({
+      userId: args.id,
+      beginAt: new Date(),
+      ...(args.until && { endAt: args.until }),
+      ...(args.reason && { reason: args.reason }),
+    })
+    await tx.transactionCommit()
+  } catch (e) {
+    await tx.transactionRollback()
+    throw e
+  }
+
+  loaders.DiscussionSuspension.clear(args.id)
+
+  return loaders.User.byId.load(args.id)
+}
--- a/packages/discussions/graphql/resolvers/_mutations/suspendUser.js
+++ b/packages/discussions/graphql/resolvers/_mutations/suspendUser.js
@@ -1,5 +1,7 @@
 const { Roles } = require('@orbiting/backend-modules-auth')
 
+const slack = require('../../../lib/slack')
+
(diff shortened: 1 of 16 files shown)


packages/discussions/migrations/sqls/20201218010032-user-suspensions-up.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE "discussionSuspensions" (
```
annatraussnig, line 1: do we also want to log who suspended that person?

packages/discussions/graphql/resolvers/_mutations/suspendUser.js
```diff
@@ -0,0 +1,25 @@
...
+  const { pgdb, loaders, user: me } = context
+
+  Roles.ensureUserIsInRoles(me, ['admin', 'supporter'])
+
+  const tx = await pgdb.transactionBegin()
+  try {
+    await pgdb.public.discussionSuspensions.insert({
+      userId: args.id,
+      beginAt: new Date(),
+      ...(args.until && { endAt: args.until }),
```
annatraussnig, line 13: Am I missing the week's default interval? I can't see it

annatraussnig (approved):

packages/discussions/graphql/resolvers/_mutations/suspendUser.js
patrickvenetz: That is [declared in database migration](https://github.com/orbiting/backends/pull/550/files#diff-c623ab3eaade6dfcf122c168b16926eca18f4fe039d29e4021e3e4640b8796ddR5). Just to have no useless records.

But a coded default, too, is more understandable, now that I think of it. 👍

packages/discussions/migrations/sqls/20201218010032-user-suspensions-up.sql
patrickvenetz: Funny, I thought so, too!

Now that we are two, I shall do so.

packages/discussions/graphql/resolvers/_mutations/suspendUser.js
annatraussnig: aaaaah, ok – you've got to have eagle eyes to spot that!! 🦅

PR #551, opened 2020-12-18
fix(publikator): Put published and re-scheduled repos in scheduled phase

patrickvenetz:
(tba)


patrickvenetz:
Closing, unwanted feature.

PR #553, opened 2020-12-22
return templateRepoId in document meta

annatraussnig:


packages/publikator/lib/cache/search.js
```diff
@@ -84,6 +84,7 @@ const find = async (args, { elastic }) => {
     'contentMeta.slug',
     'contentMeta.subject',
     'contentMeta.template',
+    'contentMeta.templateRepoId',
```
patrickvenetz, line 87: `fields` are queried for user search terms. I think querying `contentMeta.templateRepoId` is not helpful. Would drop it. (This is true for `contentMeta.template`, too.)

patrickvenetz (approved):

PR #569, opened 2021-03-02
feat(publikator): Add base migrations

patrickvenetz:
Includes schema for repos, commits and milestones, adds triggers and
alters notifyTableChangeQueue to accept ids as text (instead of uuids)

### DIFF
--- a/packages/publikator/migrations/sqls/20210201080354-base-down.sql
+++ b/packages/publikator/migrations/sqls/20210201080354-base-down.sql
@@ -6,6 +6,7 @@ TRUNCATE public."notifyTableChangeQueue";
 ALTER TABLE public."notifyTableChangeQueue"
   ALTER COLUMN "id" TYPE uuid USING id::uuid;
 DROP TABLE publikator.milestones;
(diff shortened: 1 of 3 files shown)


packages/publikator/migrations/sqls/20210201080354-base-down.sql
```diff
@@ -0,0 +1,8 @@
+DROP TRIGGER IF EXISTS trigger_publikator_commits_notify_change
+  ON publikator.commits;
+DROP TRIGGER IF EXISTS trigger_publikator_repos_notify_change
+  ON publikator.repos;
+DROP TABLE publikator.milestones;
+DROP TABLE publikator.commits;
+DROP TABLE publikator.repos;
```
fuenkchen, line 7: In up script we have `ALTER TABLE "public"."notifyTableChangeQueue" ALTER COLUMN "id" TYPE text;`, shouldn't it be changed back in down script too?

patrickvenetz: Indeed!

fuenkchen (approved):
🚢

PR #571, opened 2021-03-10
fix(mail) remove golden pin from benefactor payment and pledge mail

fuenkchen:
* golden pins are not sent via mail anymore so it's removed from mail templates


patrickvenetz (approved):
LGTM

PR #577, opened 2021-03-26
Sharegenerator

annatraussnig:
make all the various fields of this feature queriable

### DIFF
--- a/packages/documents/graphql/schema-types.js
+++ b/packages/documents/graphql/schema-types.js
@@ -50,23 +50,13 @@ type Meta {
   facebookTitle: String
   facebookImage: String
   facebookDescription: String
-  facebookGenerated: Boolean
-  facebookText: String
-  facebookFontSize: Int
-  facebookFontStyle: String
-  facebookColoredBackground: Boolean
-  facebookIlluBackground: Boolean
-  facebookTextPosition: String
   twitterTitle: String
   twitterImage: String
   twitterDescription: String
-  twitterGenerated: Boolean
-  twitterText: String
-  twitterFontSize: Int
-  twitterFontStyle: String
-  twitterColoredBackground: Boolean
-  twitterIlluBackground: Boolean
-  twitterTextPosition: String
+  shareText: String
+  shareFontSize: Int
+  shareInverted: Boolean
+  shareTextPosition: String
   prepublication: Boolean
   publishDate: DateTime
   lastPublishedAt: DateTime
@@ -78,8 +68,8 @@ type Meta {
   section: Document
   format: Document
   dossier: Document
-  shareImage: String
-  shareImageColor: String
+  shareBackgroundImage: String
+  shareBackgroundImageInverted: String
 
   credits: JSON
   audioSource: AudioSource
--- a/packages/documents/graphql/schema-types.js
+++ b/packages/documents/graphql/schema-types.js
@@ -64,6 +64,8 @@ type Meta {
   section: Document
   format: Document
   dossier: Document
+  shareImage: String
+  shareImageColor: String
 
   credits: JSON
   audioSource: AudioSource
--- a/packages/documents/graphql/schema-types.js
+++ b/packages/documents/graphql/schema-types.js
@@ -50,9 +50,23 @@ type Meta {
   facebookTitle: String
   facebookImage: String
   facebookDescription: String
+  facebookGenerated: Boolean
+  facebookText: String
+  facebookFontSize: Int
+  facebookFontStyle: String
+  facebookColoredBackground: Boolean
+  facebookIlluBackground: Boolean
+  facebookTextPosition: String
   twitterTitle: String
   twitterImage: String
   twitterDescription: String
+  twitterGenerated: Boolean
+  twitterText: String
+  twitterFontSize: Int
+  twitterFontStyle: String
+  twitterColoredBackground: Boolean
+  twitterIlluBackground: Boolean
+  twitterTextPosition: String
   prepublication: Boolean
   publishDate: DateTime
   lastPublishedAt: DateTime


packages/documents/graphql/schema-types.js
```diff
@@ -64,6 +78,8 @@ type Meta {
   section: Document
   format: Document
   dossier: Document
+  shareImage: String
+  shareImageColor: String
```
patrickvenetz, line 82: This won't produce an error but is stilled checked if it is an image URL. Maybe `shareColor`?

annatraussnig: this one is an image

annatraussnig: is the type wrong?

patrickvenetz: Oh my! I assumed `shareImageColor` is a hex color code which makes my comment obsolete.

packages/documents/graphql/schema-types.js
```diff
@@ -50,9 +50,23 @@ type Meta {
   facebookTitle: String
   facebookImage: String
   facebookDescription: String
+  facebookGenerated: Boolean
+  facebookText: String
+  facebookFontSize: Int
+  facebookFontStyle: String
```
tpreusse, line 56: (optional) Just to double check: what are the use cases for custom font styles?

Usually we infer from `meta.format.meta.kind` and `meta.kind`:
```
                  const kind =
                    (format && format.meta && format.meta.kind) ||
                    (meta && meta.kind)

                  const Headline =
                    metaHeadlines || kind === 'meta'
                      ? Meta.Headline
                      : kind === 'scribble'
                      ? Scribble.Headline
                      : Editorial.Headline
```
[Article template](https://github.com/orbiting/styleguide/blob/master/src/templates/Article/index.js#L288-L297)

packages/documents/graphql/schema-types.js
```diff
@@ -50,9 +50,23 @@ type Meta {
   facebookTitle: String
   facebookImage: String
   facebookDescription: String
+  facebookGenerated: Boolean
+  facebookText: String
```
tpreusse, line 54: (optional) Could we combine this in more prop? And maybe `facebookImageText` to be a bit more distinct?

tpreusse (commented):
I think we should do one naming and simplification iteration here

packages/documents/graphql/schema-types.js
```diff
@@ -50,9 +50,23 @@ type Meta {
...
+  facebookFontSize: Int
+  facebookFontStyle: String
+  facebookColoredBackground: Boolean
+  facebookIlluBackground: Boolean
```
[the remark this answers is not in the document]
annatraussnig: I thought that the illu was optional. @ovbm?

packages/documents/graphql/schema-types.js
annatraussnig: Yeah... that's more of a convenience/clarity prop. Otherwise I have to manually erase any text when the producer switch to a custom image upload. Might also be useful to have those 2 decoupled if we want to derive a default text from something else (e.g. title, use twitter text for facebook, etc.)

packages/documents/graphql/schema-types.js
annatraussnig: For meta share images, there are 2 different styles in use, as far as I know. We also support formatless share images, which don't have a predefined font style. Is this correct, @ovbm?

ovbm: Usecase for custom fontstyles are sharetiles like meta, where they could choose in the sketch file. 
Screenshot 2021-04-14 at 10 38 24

packages/documents/graphql/schema-types.js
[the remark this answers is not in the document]
ovbm: In sketch it's optional. let me ask production!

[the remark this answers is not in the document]
tpreusse: Oh I didn't realise that this toggles whether the background is being used. Could we rename to `facebookUseBackground` or something? This way it potentially could also be used in the future for other optional backgrounds that are not illustrated.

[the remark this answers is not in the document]
ovbm: Well they just told me they don't ever use the option witout a background image, so we can remove the toggle and infer it... (sketch spec and reality are two different things XD)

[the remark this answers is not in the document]
annatraussnig: i'll take care of it. Aiming for a mid-June release.

packages/documents/graphql/schema-types.js
ovbm: BTW: Production confirms thomas: They only use meta font for meta....

annatraussnig: Why is Thomas always right? 😸 
So custom font style would only be for format-less docs. Is this a use case we want to support or shall we drop it and only show the generator when a format is present?

tpreusse: 🙈 

I would drop it. But keep the generator for documents without format just always with Rubis. Once we have concrete use cases for switching font styles for documents without formats we decide on how to infer it (e.g. from meta.kind or meta.template or a new prop).

ovbm: I would say show it for format-less docs.

annatraussnig: well all this gives me more work BUT it also simplifies things, which is good, I guess.

ovbm: 🤪 sorry for that

packages/documents/graphql/schema-types.js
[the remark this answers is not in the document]
annatraussnig: I'll keep the prefix `share-`. As per production's feedback, we are merging facebook and twitter sharetafel props into one, meaning, e.g. `facebookText` and `twitterText`become `shareText`. Then the meta prefix is consistent.

PR #582, opened 2021-03-31
(mail) extend pledge abo template

fuenkchen:
* add pledge var: boolean if pledger already has monthly abo
* extend mail template: use another text block when user upgrades from monthly to abo


patrickvenetz (approved):
LGTM

PR #595, opened 2021-05-04
simplify discussion path logic and resolve actual path in graphql

tpreusse:
SQL Update Needed:
```
UPDATE "public"."discussions" SET "path"='/dialog?t=general' WHERE "id"='ec6ba678-8eb2-45e0-97fa-20d3a4d58b64';
```

Changes
- move general path to DB
- simplify logic and make it more performant, no more document loading
- resolve `Discussion.path` to the actual frontend path used

Screenshot 2021-05-04 at 14 42 04

### DIFF
--- a/packages/discussions/lib/Notifications.js
+++ b/packages/discussions/lib/Notifications.js
@@ -25,21 +25,21 @@ const {
   DEFAULT_MAIL_FROM_ADDRESS,
   DEFAULT_MAIL_FROM_NAME,
   FRONTEND_BASE_URL,
-  GENERAL_FEEDBACK_DISCUSSION_ID,
 } = process.env
 
-const getDiscussionUrl = async (discussion, context) => {
-  const communityUrl = `${FRONTEND_BASE_URL}/dialog?id=${discussion.id}`
-  if (discussion.id === GENERAL_FEEDBACK_DISCUSSION_ID) {
-    return `${communityUrl}&t=general`
+const getDiscussionPath = (discussion) => {
+  // currently discussion.path is synched with document path on publish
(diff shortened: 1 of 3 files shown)


packages/discussions/lib/Notifications.js
```diff
@@ -25,21 +25,21 @@ const {
...
-  const communityUrl = `${FRONTEND_BASE_URL}/dialog?id=${discussion.id}`
-  if (discussion.id === GENERAL_FEEDBACK_DISCUSSION_ID) {
-    return `${communityUrl}&t=general`
+const getDiscussionPath = (discussion) => {
+  // currently discussion.path is synched with document path on publish
+  // - for template discussion the document path is always suffixed with /diskussion
+  //   in that case we don't want to rewrite the path
+  // - in all other cases, mainly template article,
+  //   we rewrite to /dialog with the discussion id
+  if (discussion.repoId && !discussion.path.match(/\/diskussion$/)) {
```
tpreusse, line 36: After sleeping on it:
To handle the edge case of template `discussion` being rewritten to a nice path ([example](https://www.republik.ch/brigitte-meyer)):
```suggestion
  if (discussion.repoId && !discussion.path.match(/\/diskussion$/) && discussion.path.match(/^\/[0-9]{4}\/[0-9]{2}\/[0-9]{2}\//)) {
```

Or revert yo the old way: query document and check that it is template `article`. Or DB migration ;-)

fuenkchen:
We go for this quick solution for now, and plan to make a broader solution with DB migration later on.

patrickvenetz:
Icing this one for now. (Keeping branch however.)

PR #605, opened 2021-05-20
fix(statistics): Favor relative date over date range or absolute date

patrickvenetz:
[yargs breaking change](https://github.com/yargs/yargs/blob/master/CHANGELOG.md#1700-2021-05-02) applies coerce before validation, thus all date fields are set.

Fix favors relative date over range or absolut date. If not set, relative date is falsy.


patrickvenetz:
Well done, myself!

patrickvenetz:
Oh! Thank you 😊

PR #609, opened 2021-06-07
Enhance verification of credentials

fuenkchen:
* adds mutation to remove verification
* mutations to change verified flag now based on credential `id` (instead of `userId` and `description`)
* adds loader for credential list
* adds filtering of credentials by non anonymous usage


patrickvenetz (approved):
LGTM

Issue #276, opened 2019-06-24
Delete a user won't remove email address from MailChimp


patrickvenetz:
## Bug Report

It was reported that deleting a users seems to not remove but only unsubscribe a user's email address from MailChimp list.

It should delete record all together.

PR #613, opened 2021-06-18
changes to MailchimpInterface

fuenkchen:
solves #276

Changes are as follows and can be tested on love
- added new method to MailchimpInterface to permanently delete member in Mailchimp
- if a user is deleted, the corresponding email address will be deleted (instead of archived) in Mailchimp
- if two users are merged, the old email address will be archived (instead of unsubscribed) in Mailchimp
- if the email address of a user is changed, the old email address will be archived (instead of unsubscribed) in Mailchimp


patrickvenetz (approved):
LGTM

patrickvenetz (approved):
LGTMOM

PR #622, opened 2021-07-05
fix(republik-crowdfundings): Reactivate cancelled but active subscription

patrickvenetz:
This Pull Request fixes a regression introduced in #522: Mutation `reactivateMembership` did no reactive cancelled but still active subscriptions.

API version 2018-02-28 or younger require `cancel_at_period_end` parameter to reactive a cancelled (but active) subscription.

Before #522 we used API version 2017-12-14 which "[any parameter sent to the Update Subscription API would automatically reactivate a subscription.](https://stripe.com/docs/billing/subscriptions/cancel#reactivating-canceled-subscriptions)"


tpreusse (approved):
LGTM

PR #630, opened 2021-08-04
Mailbox queries, cookie SameSite=None and "pageini"

patrickvenetz:
This Pull Request adds [`mailbox` queries to root and `User.mailbox`](https://github.com/orbiting/backends/compare/pae/feat-zat?expand=1#diff-69df5e7dfca0bc0e0a1ef244a5233a9571ec27938974adbf1ee28238159cbf83) to query (new) mail index and `mailLog` table at once and thus [deprecating `mailLog` queries](https://github.com/orbiting/backends/compare/pae/feat-zat?expand=1#diff-b941647b52246b2ad69ba7b26189886545bea78af20782344e25adcbc6e7992f).

It changes [session cookie attribute `SameSite` to `None`](https://github.com/orbiting/backends/compare/pae/feat-zat?expand=1#diff-192366cb64dcd51bed8ffd63a9aaddc48d8fee59d907daa1e7f553ed9bc0abc2) if not in production. (This might pose an inadvertent security risk and we've to dig into that one deeper.)

It implements a [new pagination approach ("pageini")](https://github.com/orbiting/backends/compare/pae/feat-zat?expand=1#diff-46877d066b4f0b57ea3b90b1c5b4518946152680f44878f245c8c0e1833b6b11): more generic, memory friendly way

### DIFF
--- a/packages/utils/paginate.js
+++ b/packages/utils/paginate.js
@@ -119,6 +119,30 @@ module.exports.paginator = (args, payloadFn, nodesFn) => {
   }
 }
 
+/**
+ * Util which helps generating an GraphQL Cursor Connection. It allows to
+ * pagination results.
+ *
+ * @see https://relay.dev/graphql/connections.htm
+ *
+ * This util implementation is somewhat superior to previous pagination
+ * implementation ("paginate", "paginator") as it won't require all data and
+ * then slice it.
+ *
+ * Flow:
+ *
+ * 1) Decode after and before cursors in {args}
+ *
+ * 2) Call {countFn} with {args}, after and before cursors. Its expected to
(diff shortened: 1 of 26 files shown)


packages/utils/paginate.js
```diff
@@ -118,3 +118,30 @@ module.exports.paginator = (args, payloadFn, nodesFn) => {
     _nodes: nodes,
   }
 }
+
```
fuenkchen, line 121: maybe it would be nice to have a short explanatory comment why we have magic pageini in addition to paginate aka pageini does that, paginator does this

fuenkchen (approved):
Looks very good to me

PR #633, opened 2021-08-19
Feat election candidacy

fuenkchen:
TODOs:

- [x] clarify if and add gender and biography to candidate checks in `updateMe.js`
- [x] ~~add INSERT statement to PR for new elections (gen21p & gen21m) as soon as candidacy time range and discussionId are defined~~ there is a mutation `createElection` for that purpose 
- [x] replace dummy confirmation mail text with definite text
- [x] add translations (e.g. subject of confirmation mail)

maybe there are some more things needed to implement / finalize frontend

### DIFF
--- a/packages/voting/graphql/schema-types.js
+++ b/packages/voting/graphql/schema-types.js
@@ -146,6 +146,7 @@ type Candidacy {
 
   yearOfBirth: Int
   city: String
+  credential: Credential
 }
 
 type ElectionTurnout {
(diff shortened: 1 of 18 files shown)


packages/voting/graphql/schema-types.js
```diff
@@ -146,6 +146,7 @@ type Candidacy {
 
   yearOfBirth: Int
   city: String
+  credential: Credential
```
tpreusse, line 149: Resolving the credential does not work yet when queried via `election.candidacies` or `me.candidacies`:
Screenshot 2021-08-29 at 00 14 07

fuenkchen: Right. Done with https://github.com/orbiting/backends/pull/633/commits/f20e06fbbc919153d18b87644de06b4c825cb7e6

PR #639, opened 2021-09-07
auto sorting of discussion comments

fuenkchen:
* adds `AUTO` to `DiscussionOrder` enum in schema
* adds `resolvedOrderBy` property to `CommentConnection` in schema
* when auto sorting is requested in Discussion.comments it will check creation date of first comment:
  * if it's within the last 24 hours it will sort by `DATE`
  * if it's older than 24 hours it will sort by `VOTES`
  * and returns it as `resolvedOrderBy`

can be tested on love

### DIFF
--- a/packages/discussions/graphql/resolvers/Discussion/comments.js
+++ b/packages/discussions/graphql/resolvers/Discussion/comments.js
@@ -4,6 +4,8 @@ const { ascending, descending } = require('d3-array')
 
 const getSortKey = require('../../../lib/sortKey')
 
+const THRESHOLD_OLD_DISCUSSION_IN_MS = 1000 * 60 * 60 * 24 // 24 hours
+
 const assembleTree = (_comment, _comments) => {
   const coveredComments = []
 
@@ -250,22 +252,20 @@ module.exports = async (discussion, args, context, info) => {
     - or by votes otherwise 
     the property resolvedOrderBy is just needed when DiscussionOrder === AUTO
   */
-  let resolvedOrderBy
-  if (orderBy === 'AUTO') {
-    const sortedComments = [...comments].sort(
-      (a, b) => Date.parse(a.createdAt) - Date.parse(b.createdAt),
-    )
-    if (sortedComments.length > 0) {
-      const firstCommentCreatedAt = Date.parse(sortedComments[0].createdAt)
-      const twentyFourHoursAgo = new Date(
-        new Date().getTime() - 24 * 60 * 60 * 1000,
-      )
-      resolvedOrderBy =
-        firstCommentCreatedAt > twentyFourHoursAgo ? 'DATE' : 'VOTES'
-    } else {
-      resolvedOrderBy = 'DATE'
+  const oldestComment = comments?.reduce((oldest, current) => {
+    if (!oldest) {
+      return current
     }
-  }
+    return oldest.createdAt < current.createdAt ? oldest : current
+  }, false)
+
+  const thresholdOldDiscussion =
+    new Date().getTime() - THRESHOLD_OLD_DISCUSSION_IN_MS
+
+  const resolvedOrderBy =
+    oldestComment && oldestComment.createdAt?.getTime() < thresholdOldDiscussion
+      ? 'VOTES'
+      : 'DATE'
 
   let tree = parentId ? comments.find((c) => c.id === parentId) : {}
 
--- a/packages/discussions/graphql/resolvers/Discussion/comments.js
+++ b/packages/discussions/graphql/resolvers/Discussion/comments.js
@@ -83,10 +83,9 @@ const deepSortTree = (
         topIds && topIds.includes(comment.id)
           ? topValue
           : bubbleSort
-          ? [
-              comment[sortKey],
-              firstChild.topValue || firstChild[sortKey],
-            ].sort((a, b) => ascDesc(a, b))[0]
+          ? [comment[sortKey], firstChild.topValue || firstChild[sortKey]].sort(
+              (a, b) => ascDesc(a, b),
+            )[0]
           : null
     }
   }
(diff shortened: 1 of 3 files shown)


packages/discussions/graphql/resolvers/Discussion/comments.js
```diff
@@ -245,6 +244,29 @@ module.exports = async (discussion, args, context, info) => {
...
+  /* 
+    AUTO = comments are sorted 
+    - by date if the first comment was created in the last 24 hours 
+    - or by votes otherwise 
+    the property resolvedOrderBy is just needed when DiscussionOrder === AUTO
+  */
+  let resolvedOrderBy
+  if (orderBy === 'AUTO') {
+    const sortedComments = [...comments].sort(
+      (a, b) => Date.parse(a.createdAt) - Date.parse(b.createdAt),
```
patrickvenetz, line 256: When sorting arrays, we usually include [`d3-array`](https://github.com/orbiting/backends/blob/60c32fd1982f4bd312ba3b1eddd990a53d449173/packages/republik-crowdfundings/graphql/resolvers/Pledge.js#L2) which takes care of edge cases.

```suggestion
      (a, b) => ascending(a.createdAt, b.createdAt),
```

It's likely not to require `Date.parse` (albeit not tested).

If you're inclined to change this, make sure to import `d3-array` at top of file first.

packages/discussions/graphql/resolvers/Discussion/comments.js
```diff
@@ -245,6 +244,29 @@ module.exports = async (discussion, args, context, info) => {
...
+    )
+    if (sortedComments.length > 0) {
+      const firstCommentCreatedAt = Date.parse(sortedComments[0].createdAt)
```
patrickvenetz, line 259: Probably no need to parse `createdAt` since Pogi will return a `Date` object or soemthing with a similar signature.

packages/discussions/graphql/resolvers/Discussion/comments.js
```diff
@@ -245,6 +244,29 @@ module.exports = async (discussion, args, context, info) => {
...
+      const twentyFourHoursAgo = new Date(
+        new Date().getTime() - 24 * 60 * 60 * 1000,
```
patrickvenetz, line 261:
```suggestion
        new Date().getTime() - 1000 * 60 * 60 * 24, // 24 hours
```

Would flip multiplication (miliseconds to seconds to minutes to hours to a day) and append a comment for faster mental parsing.

packages/discussions/graphql/resolvers/Discussion/comments.js
```diff
@@ -245,6 +244,29 @@ module.exports = async (discussion, args, context, info) => {
     }
   }
 
+  /* 
+    AUTO = comments are sorted 
+    - by date if the first comment was created in the last 24 hours 
+    - or by votes otherwise 
+    the property resolvedOrderBy is just needed when DiscussionOrder === AUTO
+  */
+  let resolvedOrderBy
```
patrickvenetz, line 253:
```suggestion
  let resolvedOrderBy = 'DATE'
```

Maybe default to `DATE` here and remove else-clause further below.

packages/discussions/graphql/resolvers/Discussion/comments.js
```diff
@@ -245,6 +244,29 @@ module.exports = async (discussion, args, context, info) => {
...
+  if (orderBy === 'AUTO') {
+    const sortedComments = [...comments].sort(
```
patrickvenetz, line 255: I was thinking, maybe a more intuitive way to get to the "oldest" entry – instead of sorting – could be a reducer. Would you feel comfortable with that, or is it contrary to my believe rather unintuitive?

Here is a mock:

```js
// Keep this rather a the top of file.
const OLD_THRESHOLD_IN_MS = 1000 * 60 * 60 * 24 // 24 hours

// Then, inside if(orderBy === 'AUTO') {...

const initialComment = comments?.reduce((oldestCommentYet, currentComment) => {
  // If there was no oldestComment, set current comment as oldest
  if (!oldestCommentYet) {
    return currentComment 
  }

  // If currentComment is older than oldestCommentYet, set current as
  // oldest. Otherwise keep oldestComment.
  return currentComment.createdAt < oldestCommentYet.endDate
    ? currentComment
    : oldestCommentYet
}, false) // Start with false. If comment array is empty, false will prevail

// 
// So, up until here, either intialComment is falsy or we got the oldest comment at hand.
// 

const oldThresholdInMs = new Date().getTime() - OLD_THRESHOLD_IN_MS

// If initialComment older than oldThresholdInMs, order by VOTES.
// And if initialComment is falsy or younger, order by DATE.
const resolvedOrderBy = (initialComment?.createdAt?.getTime() < oldThresholdInMs && 'VOTES') || 'DATE'
```

patrickvenetz:
@fuenkchen Maybe start with this [comment](https://github.com/orbiting/backends/pull/639#discussion_r704377345) – other suggestions might get obsolete then.

packages/discussions/graphql/resolvers/Discussion/comments.js
fuenkchen: Yes, I agree, much more elegant and also better in terms of performance I think

packages/discussions/graphql/resolvers/Discussion/comments.js
fuenkchen: kept it undefined, because when `orderBy !== 'AUTO'` it should not return a value

patrickvenetz (changes requested):
Per PR description, I'd have expected `query comments(orderBy: AUTO) {...}` to set `resolvedOrderBy`, too. It currently returns `null`.

I'd argue, that queries with argument `orderBy` which returns `CommentConnection`, should reveil actual ordering in `resolvedOrderBy`. Thus far, that applies to `Discussion.comments` (done and all) and `comments`.

fuenkchen:
Of course, you're so right. I'll change that

patrickvenetz (approved):

Issue #504, opened 2020-09-07
Remove commented out code


caillou:
In the following commit, some code has been commented out, as the code was not fully tested and somewhat critical:
 
https://github.com/orbiting/backends/commit/eac5e533b18a7a0ae5053edf5b06c395bc49152f

Once we are sure this method is called as expected, we should un-comment the code.

PR #642, opened 2021-09-20
actually set auto pay to false if data is missing

fuenkchen:
fixes https://github.com/orbiting/backends/issues/504


patrickvenetz (approved):
LGTM

Issue #558, opened 2021-01-07
Fix DEV variables


patrickvenetz:
## Bug Report

[DEV variable is either undefined, `false` or `true`](https://github.com/orbiting/backends/blob/master/servers/graphql/server.js#L74).

```js
const DEV = NODE_ENV && NODE_ENV !== 'production'
```

If `NODE_ENV` is missing, DEV is undefined. This leads to unwanted behaviour when DEV is checked to be truthy or falsy i.e. it might run schedulers despite being in development environment.

Ensure `DEV` returns a boolean value [here](https://github.com/orbiting/backends/blob/master/servers/graphql/server.js#L74) and in multiple other files, too.

[A working example](https://github.com/orbiting/backends/blob/master/packages/env/index.js#L3):

```js
const DEV = process.env.NODE_ENV ? process.env.NODE_ENV !== 'production' : true
```

PR #643, opened 2021-09-20
It's always DEV....

fuenkchen:
... except when it's production
... then it's not dev

fixes https://github.com/orbiting/backends/issues/558


patrickvenetz (approved):
LGTM

PR #649, opened 2021-09-21
better error handling when generating invoices/paymentslips

fuenkchen:
returns error response if any error happens within `pdfReqHandler` method


patrickvenetz (approved):
LGTM

PR #650, opened 2021-09-21
remove legacy stuff in exists method of importPayments

fuenkchen:
* removes conversion to utf-8 and trimming of `avisierungstext` in `exists` method

(there is some formatting noise, bit surprised about that, but I guess it's fine)


patrickvenetz (approved):
LGTM

Issue #517, opened 2020-09-21
Notifications of failed mergeUsers()


caillou:
Notify the developers in case of a failed `mergeUsers()` migration.

PR #652, opened 2021-09-23
mergeUsers: add slack notification if anything goes wrong

fuenkchen:
solves https://github.com/orbiting/backends/issues/517


patrickvenetz (approved):
LGTM

PR #653, opened 2021-09-27
(republik) fix check in updateMe.js 

fuenkchen:
* fix check when user who is candidating for GR wants to change portrait image

can be tested on love


patrickvenetz (approved):
LGTM

PR #654, opened 2021-09-27
fix(utils): Harden against missing value

patrickvenetz:
`child.value` is not always present in MDAST when formatting a link itself. In such cases, do not collapse link.


fuenkchen (approved):
lgtm

PR #658, opened 2021-10-05
feat(Document): resolve meta seo title and seo description

tpreusse:
Publikator introduces new seo fields which also should be exposed via graphql:
https://github.com/orbiting/publikator-frontend/pull/302

Screenshot 2021-10-05 at 17 40 27

@patrickvenetz do you want to add the fields to our elastic index? Maybe we should also add twitter title and lead then? Will be used for search engines if seo fields are not filled.


patrickvenetz (approved):
LGTM

Not adding SEO and Twitter wording to search index.

PR #661, opened 2021-10-08
feat(base): Add PgDb.queryInBatches

patrickvenetz:
Allows to query postgress in batches:

```js
await pgdb.queryInBatches(
  { handleFn: (rows) => { /* Do something */ }, size: 100 }
  `SELECT * FROM users`
)
```


fuenkchen (approved):
lgtm!

PR #665, opened 2021-10-22
fix(redirections): Drop keepQuery flag

patrickvenetz:
This Pull Request removes option `keepQuery` which did forward and merge query parameters passed in `query redirection(path)` (e.g. "/some/path?foo=bar"). It will ignore passed parameters.

Other changes inlucde:
- Querying path with a hostname other than `FRONTEND_BASE_URL` won't check for a redirection
- `Redirection.get` returns only one or no record (instead of an array of records)
- Adding a redirection does not require status code anymore. It will default to 302 (Found).


fuenkchen (approved):
lgtm

PR #673, opened 2021-11-05
fix: Check overview is linked to a series master

patrickvenetz:
This Pull Request will throw an error on commit if overiew points to anything but a series master document. It also allows to search for repos on which last commit indicates a series master or episode document.

Each document in a series should point to a series master.

In a master document, `meta.series.overview` should contain a repo URL (and is usually a self-reference). In an episode document, `meta.series` should contain a repo URL.

`meta.series` is either an object (when master) or a string (when episode).

`repoSearch` can query for `isSeriesMaster` or `isSeriesEpisode`.


fuenkchen (approved):
❤️

PR #675, opened 2021-11-09
(discussion) increase threshold for resolved order of comments

fuenkchen:
* comments sorted by votes only after 72 hours not 24 hours


patrickvenetz (approved):
LGTM

PR #683, opened 2021-11-19
Introduce moderator role

patrickvenetz:
This Pull Request introduces `moderator` role:
- … on mutation `unpublishComment` (while `editor` can no longer call that mutation)
- … on `Comment.numReports` (while `editor` can no longer query that prop)
- … exposed on `User.roles`

Other changes include:
- `supporter` role can query `DiscussionSuspension.issuer` (while `editor` can no more)
- `admin` role can call `createDiscussion`
- only `admin` role can query `Comment.author`


trm217:
## Should be batch deployed with
- https://github.com/orbiting/republik-frontend/pull/688
- https://github.com/orbiting/republik-admin-frontend/pull/79

fuenkchen (approved):
👍

PR #685, opened 2021-11-22
(mail) cleaned user mail

fuenkchen:
- adds script to send mails to a certain segment of mail addresses
- adds script to get cleaned users and provide specific mail information
- adds mail template for cleaned users to invite them to get back to newsletter subscription

Possible further improvements
- move temporary templates to another location and extend `getTemplates` method so that one could specify a path
- there are some code duplicates in `sendScheduledMails.js` and `sendMailsToSegment.js` - maybe this could be generalized too
- atm one can use `envMergeVars` but cannot specify further vars which will be merged together, this could be done now or whenever it's really needed
- maybe use `queryInBatches` (or as option if number in segment is over certain threshold)

### DIFF
--- a/packages/mail/express/mailchimp.js
+++ b/packages/mail/express/mailchimp.js
@@ -1,6 +1,10 @@
 const bodyParser = require('body-parser')
 const moment = require('moment')
 
+const {
+  enforceSubscriptions,
+} = require('@orbiting/backend-modules-republik-crowdfundings/lib/Mail')
+
 const { MAIL_EXPRESS_MAILCHIMP_SECRET } = process.env
 
 module.exports = async (server, pgdb) => {
@@ -31,6 +35,7 @@ module.exports = async (server, pgdb) => {
       switch (type) {
         case 'subscribe':
           Object.assign(record, handleSubscribe(data))
+          await applyEnforceSubscriptions(record, pgdb)
           break
         case 'unsubscribe':
           Object.assign(record, handleUnsubscribe(data))
@@ -178,4 +183,14 @@ const getGroups = (name, data) => {
     .split(',')
     .map((g) => g.trim())
     .filter(Boolean)
+}
+
+const applyEnforceSubscriptions = async (record, pgdb) => {
+  try {
+    const { email } = record
+    const user = await pgdb.public.users.findOne({ email })
+    await enforceSubscriptions({ pgdb, userId: user?.id, email })
+  } catch (e) {
+    console.warn('applyEnforceSubscriptions failed:', e)
+  }
 }
--- a/packages/mail/express/mailchimp.js
+++ b/packages/mail/express/mailchimp.js
@@ -98,7 +98,7 @@ const handleUnsubscribe = (data) => {
     email: data.email,
     action: data.action,
     reason: data.reason,
-    campaign: data['campaign_id'],
+    campaign: data.campaign_id,
     customer: getGroups('Customer', data),
     newsletter: getGroups('Republik NL', data),
   }
@@ -131,8 +131,8 @@ const handleUpemail = (data) => {
   "data[old_email]": "api+old@mailchimp.com"
   */
   return {
-    email: data['new_email'],
-    oldEmail: data['old_email'],
+    email: data.new_email,
+    oldEmail: data.old_email,
   }
 }
 
@@ -146,7 +146,7 @@ const handleCleaned = (data) => {
   return {
     email: data.email,
     reason: data.reason,
-    campaign: data['campaign_id'],
+    campaign: data.campaign_id,
   }
 }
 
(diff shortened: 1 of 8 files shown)


packages/mail/script/sendMailsToSegment/cleanedUserSubscriptionInvitation.js
```diff
@@ -0,0 +1,68 @@
...
+      JOIN memberships m 
+        ON m."userId" = u.id AND m."active" = TRUE
+      WHERE 
+        type IN ('cleaned')
+      ORDER BY data."createdAt" DESC
+    )
+    
+    SELECT email 
+    FROM records 
+    WHERE "createdAt" >= now() - '60 days'::interval
```
patrickvenetz, line 56: Since wording in email says, that you haven't heard from us in a while, would extend to range, say, last 30 to 60 days (instead of now to 60 days).

And from-to-range could be arguments instead of limiting rows.

packages/mail/templates/cleaned_user_subscription_invitation.html
```diff
@@ -0,0 +1,155 @@
...
+                      auf die Inhalte zugreifen und die
+                      <a href="{{frontend_base_url}}/newsletter">Newsletter</a>
+                      Ihr Postfach eigentlich nur überflutet haben. Dann müssen
+                      Sie jetzt auch nichts weiter unternehmen und dürfen sich
+                      gerne wieder der Lektüre von Interessanterem als dieser
+                      Nachricht widmen. Bitte entschuldigen Sie die Störung.
+                    </p>
+                    <p>
+                      PPS: Kennen Sie schon unsere
+                      <a href="{{frontend_base_url}}/gebrauchsanleitung"
```
patrickvenetz, line 114: Shorthand link from merge variables: `{{link_manual}}`

packages/mail/script/sendMailsToSegment/cleanedUserSubscriptionInvitation.js
```diff
@@ -0,0 +1,68 @@
...
+    default: true,
+  })
+  .option('limit', {
+    default: 1000,
+  })
+  .help()
+  .version().argv
+
+PgDb.connect().then(async (pgdb) => {
+  if (argv.dryRun) {
```
patrickvenetz, line 17: This flag is not doing what description says; it will exit before able to schedule email (instead of sending).

packages/mail/templates/cleaned_user_subscription_invitation.html
fuenkchen: Ah yes, searched for gebrauchsanleitung and didn't find it

packages/mail/templates/cleaned_user_subscription_invitation.html
```diff
@@ -0,0 +1,155 @@
...
+                      border="0"
+                      cellspacing="0"
+                      cellpadding="0"
+                    >
+                      <tr>
+                        <td>
+                          <table border="0" cellspacing="0" cellpadding="0">
+                            <tr>
+                              <td align="center" bgcolor="#3CAD00">
+                                <!-- TODO Link zum subscribe formular !!!! -->
```
patrickvenetz, line 72:
```suggestion
```

packages/mail/script/sendMailsToSegment/cleanedUserSubscriptionInvitation.js
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  const emailAddresses = await pgdb.queryOneColumn(
+    `
+    WITH records AS (
+      WITH data AS (
+        SELECT
+          DISTINCT ON ("email")
+          "email",
+          "createdAt",
```
patrickvenetz, line 41: Found a flaw in original query: Better to query `firedAt` throughout instead of `createdAt`. Latter is time of insertion into database, and that might lag behind if webhook flow is broken.

packages/mail/script/sendMailsToSegment/cleanedUserSubscriptionInvitation.js
```diff
@@ -0,0 +1,77 @@
...
+          EXTRACT(DAYS FROM now() - "createdAt") days,
```
patrickvenetz, line 42: (Not used later.)

```suggestion
```

patrickvenetz (approved):
LGTM

packages/mail/express/mailchimp.js
```diff
@@ -179,3 +184,13 @@ const getGroups = (name, data) => {
     .map((g) => g.trim())
     .filter(Boolean)
 }
+
+const applyEnforceSubscriptions = async (record, pgdb) => {
+  try {
+    const { email } = record
+    const user = await pgdb.public.users.findOne({ email })
```
fuenkchen, line 191: Is there any specific reason why not using `findOneFieldOnly` to get userId directly

patrickvenetz: No, indeed, this could be `findOneFieldOnly`.

PR #694, opened 2021-12-13
refactor(republik-crowdfundings): Use PgDb.queryInBatches

patrickvenetz:
This Pull Request uses `PgDb.queryInBatches` to speed up owners scheduler and reduce memory und CPU load.


fuenkchen (approved):
lgtm 🤓

PR #696, opened 2021-12-14
feat(auth): Sunset User.eventLog

patrickvenetz:
- `User.eventLog` returns an empty array
- Database triggers to update eventLog removed


fuenkchen (approved):
👍

PR #698, opened 2021-12-15
fix(push-notifications): Split tokens into chunks of 1000 tokens

patrickvenetz:
`sendToDevice` accepts up to 1000 tokens per call


fuenkchen (approved):
👍

PR #703, opened 2021-12-24
feat(votings): Add flag VotingInterface.requireAddress

fuenkchen:
Implements flag `VotingInterface.requireAddress` in types `Election`, `Voting`.


patrickvenetz (approved):
LGTM

packages/documents/graphql/schema-types.js
```diff
@@ -95,6 +95,21 @@ type Meta {
...
+
+type TagMapping {
+  tag: String!
+  text: String!
+  color: TagMappingColor
+}
+
+type TagMappingColor {
+  light: String!
+  dark: String!
```
tpreusse, line 112: IMO we should remove this here again. I can get it from `content.meta` without it being in the schema. If we end up using it again we should move it to the discussion type.

patrickvenetz: Reverted.

PR #705, opened 2022-01-06
fix(voting): Require allowed membership to be active

patrickvenetz:
This Pull Request will require either of allowed memberships on a user to be active.

If not, a user with an inactive membership – either used or new – might be eligable to submit a ballot or answer.

### DIFF
--- a/packages/voting/lib/queries.js
+++ b/packages/voting/lib/queries.js
@@ -225,7 +225,7 @@ const buildQueries = (tableName) => {
 
     const where = [
       allowedRoles && 'u.roles ?| :roles',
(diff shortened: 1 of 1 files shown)


tpreusse:
A random thought on this, I think the reason that they didn't have to be active in the past is that:
1. for ur-abst only the time the vote starts counts – no longer needs to be active when voting
2. the number of eligible stays the same even when someones membership runs out

But is this a non issue since ur-abst are always done through roles nowadays?

patrickvenetz:
Yes, we pin Ur-Abst to roles. It proved to be more stable than pinning it to membership types (cancelling pledges still lead to unwanted diffs).

PR #706, opened 2022-01-08
fix(publikator): Handle redirection upsert sequentially

patrickvenetz:
Upserting redirections while publishing let to multiple rows with same `source`. To prevent, this Pull Request switches to `Promise.each` instead of `Promise.all`, and it uniquifies previous paths before upserting.

Adds `findPublications`, which returns all "publication" docs stored in ElasticSearch. These are docs that were once published or scheduled to be published.

We want to generate a redirection record for each publication, since a path in a scheduled publication might have been hardcoded elswhere (e.g. a newsletter) while a path might change again before a scheduled publication is actually published.

(It reverts `findPublished`, as it should return actually published or currently scheduled docs.)

### DIFF
--- a/packages/publikator/lib/Document.js
+++ b/packages/publikator/lib/Document.js
@@ -25,7 +25,7 @@ const getPath = ({ slug, template, publishDate, prepublication, path }) => {
       !!prepublication && PREFIX_PREPUBLICATION_PATH,
       ...path.split('/'),
     ]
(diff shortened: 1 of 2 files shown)


patrickvenetz:
(@fuenkchen This is a more stable version of https://github.com/orbiting/backends/pull/702.)

PR #708, opened 2022-01-11
chore: Unused indices in Postgres, ElasticSearch

patrickvenetz:
`heroku pg:diagnose` reports unused Postgres indices `notifications::notifications_event_id_idx` and `mailLog::mail_log_email_lower_idx`. In ElasticSearch mappings on documents, comment and users, some more indices are never used either.

This Pull Request drops these mercilessly.

Other changes include:
- Prefix to path for prepublished documents is exposed as ENV variable `PREFIX_PREPUBLICATION_PATH` (instead of hardcoded).
- For development purposes, one can suppresses downloading and measuring audio files while indexing a document with ENV variable`SUPPRESS_AUDIO_DURATION_MEASURE`.
- Due to licensing changes, ElasticSearch is not publishing an OSS version anymore. 
- Fixes missing Embed dataloader when indexing document zones

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -252,6 +252,9 @@ GITHUB_LOGIN=
 # maintenance: disable publishing
 # DISABLE_PUBLISH=true
 
+# Prefix to path for prepublished documents
+PREFIX_PREPUBLICATION_PATH=vorschau
+
 #############
 # republik
 #############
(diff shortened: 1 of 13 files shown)


fuenkchen (approved):
LGTM

PR #710, opened 2022-01-12
chore(republik-crowdfundings): Send payment reminders at 10pm, too

patrickvenetz:
This Pull Request adds an additional scheduler to send payment reminders at 10pm.

(Other schedulers send reminders at 4am and 12pm.)

To send payment reminders, all payment need to be "matched" to pledge payment.

Now and then, Finance is matching payments late in the afternoon (a manual stop necessary if automatic matching fails). Yet, importing payments at 4am might lead to unmatched payments again, pushing sending reminders further off again.

### DIFF
--- a/packages/republik-crowdfundings/lib/scheduler/index.js
+++ b/packages/republik-crowdfundings/lib/scheduler/index.js
@@ -46,11 +46,23 @@ const init = async (context) => {
 
   schedulers.push(
     timeScheduler.init({
-      name: 'payment-reminders',
+      name: 'payment-reminders-lunch',
       context,
       runFunc: (_args, context) => sendPaymentReminders(context),
       lockTtlSecs,
       runAtTime: '12:00',
+      runAtDaysOfWeek: [1, 2, 3, 4, 5], // Send reminder Monday to Friday
+    }),
+  )
+
+  schedulers.push(
+    timeScheduler.init({
+      name: 'payment-reminders-nightshift',
+      context,
+      runFunc: (_args, context) => sendPaymentReminders(context),
+      lockTtlSecs,
+      runAtTime: '22:00',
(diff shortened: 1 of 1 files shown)


tpreusse:
do we really want to remind people to pay us money late at night? I'd vote for 18:00 as alt slot and if our peeps work later it just doesn't immediately go out.

patrickvenetz:
Time doesn't matter IMO; we remind them at 4am, too.

tpreusse:
I think time matters. 4am is okey because normal people are asleep and read it in the morning. Honestly conceptually I think the best would be to not introduce a new time at all. But if they really want it – a reminder based on outdated old data, earlier is better.

patrickvenetz:
Na, pretty sure it has no effect. How about we measure whether time of day matters?

> But if they really want it – a reminder based on outdated old data, earlier is better.

Not sure I understand this one. «Send mail whether or not matching is complete», is that what you mean? (Which is in blatant 180 to not sending reminders until matching is complete, when you introduce automation over a year ago.)

tpreusse:
> Not sure I understand this one. «Send mail whether or not matching is complete», is that what you mean? (Which is in blatant 180 to not sending reminders until matching is complete, when you introduce automation over a year ago.)

No what I'm saying is that:
1. bank data comes in once a day, it's ready around 1am
2. people are usually awake between 5am and 11pm, in that time frame they do their banking

The ideal time to send reminder is therefore before 5am and after 1am. This is to minimize reminders being sent while transaction might already be incoming.

The noon sending time was a compromise for our accounting to not work between 1am and 5am – which would be unnecessary. So we added an extra sending time at noon. With the risk that transactions made by humans between 5am and noon are not yet imported. If the accounting department now says that they can't work in the morning I'm finde with extending the window. But not beyond 6pm – the data is already super old at that point – no new data has come in for over 15 hours. Sending reminders based on outdated data is bad.

PR #711, opened 2022-01-13
feat(auth): Post slack message on addUserToRole, removeUserFromRole

patrickvenetz:
Posts a slack message if mutations `addUserToRole` or `removeUserFromRole` are called.

It also updates `transformUser`: If `_raw` is available, user object has been transformed already and returns early with already transformed user.


fuenkchen (approved):
LGTM

PR #713, opened 2022-01-14
chore(maillog): Say good-bye to packages/maillog 👋

patrickvenetz:
This Pull Request drops "maillog" package. It removes GraphQL schema, resolvers, loaders and libs, moves migrations and express routes, a Mandrill webhook, to "mail" package.

(`mailLog` queries were flagged as deprected for a while. It's suggested to use `mailbox` instead.)

⚠️  Mandrill webhook endpoint changes to `/mail/mandrill/webhook` (instead of `/maillog/mandrill/webhook`)


fuenkchen (approved):
Goodbye beloved maillog-package, you'll stay in our hearts 👋

PR #714, opened 2022-01-15
feat(republik-crowdfundings): Put concurrency number in ENV var

patrickvenetz:
This Pull Request allows to set concurrency number on membership-owners scheduler, which describes users handled at the same time. It defaults to 10.


fuenkchen (approved):
Bueno

